PR #3, opened 2021-02-26
Move to apischema

thomascobb:
I reckon this is ready to merge now, what do you think @EmmaCD ?

We can switch to using Annotated for field and other tidying up later...


EmmaCD:
Yes agreed

Issue #6, opened 2021-03-08
Add __all__ to all the public modules


thomascobb:
Anything that appears in the docs should be in the `__all__` for its module. This should appear at the top of the module.
For instance in spec.py we should have at least:
```python
__all__ = [
   ...
    "Line",
    "Spiral",
    ...
]
```

MichaelStubbings:
In line with pull request: https://github.com/dls-controls/scanspec/pull/17 I have created '__all__' lists in every relevant source file determined by the API section of the documentation. I suggest that the issue be reviewed in the next version of the pull request and closed if satisfactory.

PR #17, opened 2021-05-04
0.3

MichaelStubbings:
- This version primarily defines a new 'getPoints' query type that can be used to return a list of plottable/scannable points from GraphQL.
- 'getPoints' returns a data structure of type 'pointsRequest' that includes multiple fields that directly relate to fields within GDA. (the current breakdown of this data structure can be seen here: https://confluence.diamond.ac.uk/display/SSCC/Data+Structure+and+Naming+Conventions)
- A new naming standard (which can also be found in the link above) has been defined.
- All relevant variable names, comments and documentation have been updated to reflect the new naming standard.
- Unit tests added to 'test_service.py' for all newly constructed code


Pipfile
```diff
@@ -24,6 +24,7 @@ sphinx-multiversion = {editable = true,git = "https://github.com/dls-controls/sp
 scanspec = {editable = true, extras = ["plotting", "graphql"], path = "."}
 # Force pipenv to install typing-extensions always (otherwise it skips on 3.8+)
 typing-extensions = "*"
+ipykernel = "*"
```
thomascobb, line 27: What is this needed for?

docs/explanations/what-are-dimensions.rst
```diff
@@ -22,34 +22,34 @@ An Example
...
-each point in the outer dimension, 18 positions in all.
+each point in the outer dimension, 18 points in all.
 
 Why a list of Dimensions?
 -------------------------
 
 A list of Dimensions are created to give the most compact representation of the
 scan. Imagine a 100x2000x2000 point scan, which creates a list of 3 Dimensions.
-Considering just the position arrays, they would take 100 + 2000 + 2000 float64
+Considering just the middle arrays, they would take 100 + 2000 + 2000 float64
```
thomascobb, line 40: I want to write "midpoint array" here...

docs/how-to/iterate-a-spec.rst
```diff
@@ -4,14 +4,14 @@ How to Iterate a Spec
 =====================
 
 A Spec is only the specification for a scan. To execute the scan we need the
-positions. We can do this in a few ways.
+frames. We can do this in a few ways.
 
-If you only need the positions
-------------------------------
+If you only need the middle points
```
thomascobb, line 9: And midpoints here...

docs/tutorials/creating-a-spec.rst
```diff
@@ -84,7 +84,7 @@ our grid we get:
 Masking with Regions
 --------------------
 
-We can apply a `Mask` to only include positions within a given `Region` using
+We can apply a `Mask` to only include points within a given `Region` using
```
thomascobb, line 87: And here too. Maybe we should rename middle -> midpoints?

scanspec/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
thomascobb, line 1: What is this file for? I thought only the .vscode in the root was used?

scanspec/cli.py
```diff
@@ -48,3 +48,11 @@ def service(cors):
     from scanspec.service import run_app
 
     run_app(cors)
+
+
+@cli.command()
+def schema():
```
thomascobb, line 54: Can we have a test for this please? Doesn't need to be very extensive, just check it prints something that is valid JSON with a "$schema" key

thomascobb: Here's some docs: https://click.palletsprojects.com/en/7.x/testing/

scanspec/regions.py
```diff
@@ -1,29 +1,29 @@
...
 
 @dataclass
 class Region(Serializable):
     """Abstract baseclass for a Region that can `Mask` a `Spec`. Supports operators:
 
-    - ``|``: `UnionOf` two Regions, positions present in either
-    - ``&``: `IntersectionOf` two Regions, positions present in both
-    - ``-``: `DifferenceOf` two Regions, positions present in first not second
-    - ``^``: `SymmetricDifferenceOf` two Regions, positions present in one not both
+    - ``|``: `UnionOf` two Regions, points present in either
```
thomascobb, line 14: Regions filter only on midpoints. This rename has shown we need to make this clear throughout...

scanspec/service.py
```diff
@@ -1,19 +1,114 @@
...
+            return None
+        else:
+            r = dtype(self._points[0])
+            s = base64.decodebytes(base64.b64encode(self._points.tobytes()))
+            t = frombuffer(s, dtype=r)
+            return array2string(t)
+
+
+@dataclass
+class axisFrames:
```
thomascobb, line 58:
```suggestion
class AxisFrames:
```

scanspec/core.py
```diff
@@ -158,9 +160,9 @@ def deserialize(cls: Type[T], serialization: Mapping[str, Any]) -> T:
         return deserialize(cls, serialization)
 
 
-#: Map of positions keys to positions_ndarray
+#: Map of axes to points_ndarray
 #: E.g. {xmotor: array([0, 1, 2]), ymotor: array([2, 2, 2])}
-Positions = Dict[Any, np.ndarray]
+Points = Dict[Any, np.ndarray]
```
thomascobb, line 165:
```suggestion
AxesPoints = Dict[Any, np.ndarray]
```

scanspec/service.py
```diff
@@ -1,19 +1,114 @@
...
+    axis: str
+    lower: Optional[Points]
+    middle: Optional[Points]
+    upper: Optional[Points]
+
+
+# The highest query level
+@dataclass
+class pointsRequest:
```
thomascobb, line 67:
```suggestion
class PointsRequest:
```

scanspec/service.py
thomascobb: Annoyingly there are 3 naming conventions for Python:
- `functions_and_methods_are_snake_case()`
- `class ClassNamesAreInitialCapsCamel:`
- `CONSTANTS_ARE_ALL_CAPS=1`

scanspec/specs.py
```diff
@@ -75,13 +75,13 @@ class Product(Spec):
         spec = Line("y", 1, 2, 3) * Line("x", 3, 4, 12)
```
thomascobb, line 75: Should turn `positions()` into `midpoints()`

thomascobb: And `SpecPositions` to `Midpoints`

scanspec/specs.py
```diff
@@ -2,7 +2,7 @@
 from typing import Any, Callable, Dict, List, Optional, TypeVar
 
 import numpy as np
-from apischema import schema
+from apischema import schema as api_schema
```
thomascobb, line 5: Why this rename? I'd prefer to keep this as its original name if possible and rename the other thing called `schema`

scanspec/specs.py
```diff
@@ -385,23 +389,25 @@ def _line_from_indexes(self, indexes: np.ndarray) -> Dict[Any, np.ndarray]:
...
     def create_dimensions(self, bounds=True, nested=False) -> List[Dimension]:
         return _dimensions_from_indexes(
-            self._line_from_indexes, self.keys(), self.num, bounds
+            self._line_from_indexes, self.axes(), self.num, bounds
         )
 
     @alternative_constructor
     def bounded(
-        key: Annotated[Any, schema(description="An identifier for what to move")],
+        axis: Annotated[Any, api_schema(description="An identifier for what to move")],
```
thomascobb, line 401: I think we should turn all `axis` types to `str`

thomascobb (changes requested):

Pipfile
MichaelStubbings: Looks like it was accidentally imported when using jupyter notebook, have since removed this and updated the pipfile.lock.

docs/tutorials/creating-a-spec.rst
MichaelStubbings: As discussed in the meeting, all instances of 'positions'/'middle' have been updated to 'midpoints'. 
Half tempted to change the lower and upper terms to lower_bounds having read through the code. But I'll let you review it first in case its okay as it is.

scanspec/.vscode/settings.json
MichaelStubbings: This file shouldnt be here, its a duplicate. I have removed it.

scanspec/specs.py
MichaelStubbings: Updated as suggested

scanspec/service.py
MichaelStubbings: Understood, made a mental (and physical) note of this

Name updated as suggested

scanspec/service.py
MichaelStubbings: Changed as requested

scanspec/regions.py
MichaelStubbings: I did wonder about this one, hopefully I have updated it to be a bit clearer

scanspec/core.py
MichaelStubbings: Suggestion to call this ScanPoints instead?

scanspec/specs.py
MichaelStubbings: https://github.com/dls-controls/scanspec/issues/6 identified as a resolution to this

scanspec/core.py
thomascobb: I think I prefer AxesPoints as this is a dict {axis: points}

scanspec/specs.py
MichaelStubbings: Agreed, all instances of axis have been updated and have had their unit tests adapted to match

scanspec/specs.py
MichaelStubbings: Added notes to the issue. Upon creating the import lists, there was no longer a conflict between the cli schema command and the imports. As such, I have reverted the naming as suggested.

scanspec/core.py
MichaelStubbings: Very well, changes made as suggested

scanspec/cli.py
MichaelStubbings: Unfortunately the schema is not stored as JSON but in the GraphQL Schema Definition Language (SDL). So we've added a couple of tests to first make sure something is printed, and then make sure that the thing that was printed is of type GraphQLSchema.

EmmaCD:
Could you add an example in docs/tutorials/graphql-service.rst for numPoints? I know lots will be added in the near future but would be nice to have the example in there to add to later.

scanspec/service.py
```diff
@@ -1,19 +1,114 @@
...
+    for axis in chunk.midpoints:
+        # Extract the midpoints and the upper and lower bounds
+        a = AxisFrames(
+            axis,
+            Points(chunk.lower.get(axis)),
+            Points(chunk.midpoints.get(axis)),
+            Points(chunk.upper.get(axis)),
+        )
+        # Append the information as a list of frames per axis
+        scan_points.append(a)
```
EmmaCD, line 101: I think with nice black formatting list comprehension would be more readable here. Also suggest getting axes from `specs.axes()` because it seems more obvious that you are getting the axes than `axis in chunk.midpoints`
```suggestion
    scan_points = [
        AxisFrames(
            axis,
            Points(chunk.lower.get(axis)),
            Points(chunk.midpoints.get(axis)),
            Points(chunk.upper.get(axis)),
        )
        for axis in spec.axes()
    ]
```

scanspec/service.py
```diff
@@ -1,19 +1,114 @@
...
-from apischema.graphql import graphql_schema
+from apischema.graphql import graphql_schema, resolver
 from graphql_server.aiohttp.graphqlview import GraphQLView, _asyncify
+from numpy import array2string, dtype, float64, frombuffer, ndarray
 
+from scanspec.core import Path
 from scanspec.specs import Spec
 
+# See confluence page for clarity on naming conventions and data structure:
+# https://confluence.diamond.ac.uk/display/SSCC/Data+Structure+and+Naming+Conventions
```
EmmaCD, line 16: Can the outside world see confluence? Is there a description of naming conventions somewhere in the docs here?

MichaelStubbings: No, only JIRA is publicly viewable. There's an explanation in the docs about what dimensions are but nothing more. I could look into adding a bit of a glossary if that would be beneficial?

scanspec/service.py
MichaelStubbings: Its a bit harder to understand for someone who's not very familiar with python, but I agree it looks much tidier. 

I have updated this in the next revision

scanspec/service.py
thomascobb: I think that would be good. This probably goes in the "Explanations" side of the docs, but with glossary definitions put in the API (by putting them in the docstrings of the classes)

docs/explanations/technical-terms
```diff
@@ -0,0 +1,27 @@
+.. _technical-terms:
+
+technical terms
```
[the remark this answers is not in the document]
thomascobb: I think the best thing to do is to link the name to the class, but keep the one line description here. The docstring is likely to have more details in it then we need.

e.g.
```
`Spec <scanspec.spec.Spec>`: A serializable representation of the type, parameters and axis names
required to produce one or more Dimensions.
```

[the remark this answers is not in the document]
thomascobb: The format is:
```
`Definition <scanspec.module.ClassName>`: description
```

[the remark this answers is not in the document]
MichaelStubbings: can you go below class level? i.e. <scanspec.core.Spec.axes>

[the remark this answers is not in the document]
thomascobb: Yes, anything that has a docs reference will work.

But isn't it scanspec.specs.Spec.axes?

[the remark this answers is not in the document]
MichaelStubbings: Yes, I'll double check these links

docs/tutorials/creating-a-spec.rst
thomascobb:
```suggestion
We can apply a `Mask` to only include frames where the midpoints are within a given `Region` using
```

docs/explanations/what-are-dimensions.rst
```diff
@@ -3,11 +3,11 @@
 What are Dimensions?
 ====================
 
-If a Spec tells you the parameters of a scan, Dimensions gives you the positions
+If a Spec tells you the parameters of a scan, Dimensions gives you the points
 that will let you actually exectute the scan. A list of Dimensions is
 interpreted as nested from slowest moving to fastest moving, so each faster
 Dimension will iterate once per position of the slower Dimension. When
-fly-scanning they key will traverse lower-position-upper on the fastest
+fly-scanning the axis will traverse lower-midpoints-upper on the fastest
```
thomascobb, line 10:
```suggestion
fly-scanning the axis will traverse lower-midpoint-upper on the fastest
```

docs/how-to/serialize-a-spec.rst
```diff
@@ -20,24 +20,24 @@ How to Serialize
...
+Product(outer=Line(axis='y', start=4.0, stop=5.0, num=6), inner=Line(axis='x', start=1.0, stop=2.0, num=3))
 
 
 How to Specify JSON from the class definitions
 ----------------------------------------------
 
 Every Spec lists in its documentation a list of parameters and types. The JSON
 representation is the dictionary of these parameters, wrapped in a dictionary
-with a single element whose key is the classname. This allows the
+with a single element whose axis is the classname. This allows the
```
thomascobb, line 40:
```suggestion
with a single element whose key is the classname. This allows the
```
This is actually a dict key rather than a scan axis!

docs/tutorials/graphql-service.rst
```diff
@@ -78,7 +78,88 @@ The equivalent in our service is:
...
+
+The 'getPoints' query makes use of this, giving users the ability to select from
+one or more of the following fields:
+
+- numFrames: the total number of frames produced by the Spec
+- returnedFrames (WIP): the number of frames returned, limited by the maxPoint arguement
+- axes: a list of axes present in the Spec and its associated scan points
+
+Within axes:
+- axis: a list of the axes present in the Spec
```
thomascobb, line 93: Isn't this the name of the axis being examined?

docs/tutorials/graphql-service.rst
```diff
@@ -78,7 +78,88 @@ The equivalent in our service is:
...
+- lower: a list of lower bounds that are each present in a frame
+- midpoints: a list of midpoints that are each present in a frame
+- upper: a list of upper bounds that are each present in a frame
+
+Within lower, middle and upper:
+- string: returns the requested points as a string
```
thomascobb, line 99: Probably best to say this is the numpy formatted string, which truncates, so is a human readable summary rather than machine readable

docs/tutorials/graphql-service.rst
```diff
@@ -78,7 +78,88 @@ The equivalent in our service is:
...
+- floatList: returns the requested points as a list of floatList
```
thomascobb, line 100:
```suggestion
- floatList: returns the requested points as a list of floats
```

tests/test_specs.py
```diff
@@ -17,67 +17,67 @@
 
 
 def test_one_point_line() -> None:
-    x = object()
+    x = "x"
```
thomascobb, line 20: Can we move this out as a global for all tests please?

```python
x = "x"

def test_something() -> None:
    inst = Line(x, 0, 1, 2)
    ...

def test_something_else() -> None:
    inst = Line(x, 2, 5, 2)
    ...
```

tests/test_specs.py
```diff
@@ -176,29 +176,29 @@ def test_squashed_product() -> None:
 
 
 def test_squashed_multiplied_snake_scan() -> None:
-    x, y, z = object(), object(), object()
+    x, y, z = "x", "y", "z"
```
thomascobb, line 179: And the same these too?

thomascobb (changes requested):
Almost there...

docs/explanations/what-are-dimensions.rst
MichaelStubbings: Updated. Also found and resolved a duplication of lower-midpoints-upper in technical terms and core.py

tests/test_specs.py
MichaelStubbings: updated

tests/test_specs.py
MichaelStubbings: updated

docs/tutorials/graphql-service.rst
MichaelStubbings: updated to "returns the requested points as a human readable numpy formatted string"

docs/tutorials/graphql-service.rst
MichaelStubbings: Currently theres no way to filter by axis, so calling axis will always return all of the axes present in the scan. Similarly, lower, midpoints and upper will always return the points present in all axes.

thomascobb: I meant that within axes, axis is the name of the axis present in the Spec, not a list

MichaelStubbings: ah I understand now

scanspec/specs.py
```diff
@@ -466,10 +487,10 @@ class Spiral(Spec):
         spec = Spiral("x", "y", 1, 5, 10, 50, 30)
     """
 
-    x_key: Any = field(
+    x_axis_name: str = field(
```
thomascobb, line 490: For consistency with Line, should we call this x_axis rather than x_axis_name?

scanspec/regions.py
```diff
@@ -185,11 +201,11 @@ class Rectangle(Region):
         spec = grid & Rectangle("x", "y", 0, 1.1, 1.5, 2.1, 30)
     """
 
-    x_key: Any = field(
-        metadata=schema(description="The key matching the x axis of the spec")
+    x_axis_name: str = field(
```
thomascobb, line 204:
```suggestion
    x_axis: str = field(
```
And the same here?

scanspec/service.py
```diff
@@ -88,19 +88,33 @@ def validate_spec(spec: Spec) -> Any:
...
     dims = spec.create_dimensions()  # Grab dimensions from spec
     path = Path(dims)  # Convert to a path
     num_points = len(path)  # Capture the length of the path
 
-    # WARNING: path object is consumed after this line
-    chunk = path.consume()
+    # Limit the consumed data to the max_points arguement
+    # # WARNING: path object is consumed after this statement
+    if max_points is None:
+        raise ValueError("max_points was set to None")
```
EmmaCD, line 111: Do we need to do this if there's a default value for max_points?

scanspec/service.py
```diff
@@ -88,19 +88,33 @@ def validate_spec(spec: Spec) -> Any:
...
+            [spec]: [The specification of the scan]
+            [max_points]: [The maximum number of points the user wishes to receive]
+
+    Returns:
+        [PointsRequest]: [A dataclass containing information about the scan points
+                            present in the spec]
     """
     dims = spec.create_dimensions()  # Grab dimensions from spec
     path = Path(dims)  # Convert to a path
     num_points = len(path)  # Capture the length of the path
```
EmmaCD, line 106: Could this be total_points or something now to avoid any ambiguity? Maybe that also requires a name change in PointsRequest? And also use this in line 112 and 113

scanspec/service.py
```diff
@@ -88,19 +88,33 @@ def validate_spec(spec: Spec) -> Any:
...
 
-    # WARNING: path object is consumed after this line
-    chunk = path.consume()
+    # Limit the consumed data to the max_points arguement
```
EmmaCD, line 108: Tiny typo

scanspec/service.py
```diff
@@ -1,19 +1,139 @@
...
+    lower: Optional[Points]
+    """The lower bounds of each midpoint (used when fly scanning)"""
+    midpoints: Optional[Points]
+    """The centre points of the scan"""
+    upper: Optional[Points]
+    """The upper bounds of each midpoint (used when fly scanning)"""
+
+
+@dataclass
+class PointsRequest:
```
callumforrester, line 74: This is the output of `getPoints` so should it be called `PointsResponse` instead?

scanspec/service.py
EmmaCD: OK I see your point here. Leave out maxPoints->max_points=200000 whereas maxPoints:null->max_points=None.

MichaelStubbings: As discussed this is a mypy problem. Without explicitly stating that an optional type may have a null value, it will complain

scanspec/service.py
MichaelStubbings: Currently this is actually the total number of frames returned so, I think total_frames would be better suited here

MichaelStubbings: On that same line of thinking, isn't max_points technically max_frames as well?

scanspec/service.py
MichaelStubbings: Yes, this makes more sense

scanspec/service.py
```diff
@@ -1,19 +1,148 @@
...
+
+# Define the schema
+schema = graphql_schema(query=[validate_spec, get_points])
+
+
+def schema_text() -> str:
+    return graphql.utilities.print_schema(schema)
 
 
 def run_app(cors=False):
```
callumforrester, line 148: For compatibility with the GDA ADE, the service needs to be able to run on a port other than 8080.
You can get a custom port as an argument in cli.py and pass it to this function. It then needs to be passed to the inner `run_app` function e.g.
```python
web.run_app(app, port=port)
```

MichaelStubbings: Added in this functionality. Can be called with scanspec service --port <port number>

PR #19, opened 2021-06-01
Regions dev

MichaelStubbings:
Adds the Ellipse and Polygon regions to scanspec

### DIFF
--- a/scanspec/regions.py
+++ b/scanspec/regions.py
@@ -260,10 +260,14 @@ class Polygon(Region):
         metadata=schema(description="The name matching the y axis of the spec")
     )
     x_verts: np.ndarray = field(
-        metadata=schema(description="The Nx1 x coordinates of the polygons vertices")
+        metadata=schema(
+            description="The Nx1 x coordinates of the polygons vertices", min_len=4
+        )
     )
     y_verts: np.ndarray = field(
-        metadata=schema(description="The Nx1 y coordinates of the polygons vertices")
+        metadata=schema(
+            description="The Nx1 y coordinates of the polygons vertices", min_len=4
+        )
     )
 
     def axis_sets(self) -> List[Set[str]]:
@@ -312,7 +316,9 @@ class Circle(Region):
     y_middle: float = field(
         metadata=schema(description="The central y point of the circle")
     )
-    radius: float = field(metadata=schema(description="Radius of the circle"))
+    radius: float = field(
+        metadata=schema(description="Radius of the circle", exc_min=0)
+    )
 
     def axis_sets(self) -> List[Set[str]]:
         return [{self.x_axis, self.y_axis}]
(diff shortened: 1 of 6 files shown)


scanspec/regions.py
```diff
@@ -242,6 +242,55 @@ def mask(self, points: AxesPoints) -> np.ndarray:
         return mask_x & mask_y
 
 
+@dataclass
+class Polygon(Region):
```
EmmaCD, line 246: Do any users ever want regular polygons or is the only use case me wanting to test it without working out vertex coordinates?

EmmaCD:
@MichaelStubbings I would appreciate an example of region masking in the grahql tutorial so I have a starting point for a query as they can be a bit long winded to write out from scratch.

scanspec/regions.py
```diff
@@ -242,6 +242,55 @@ def mask(self, points: AxesPoints) -> np.ndarray:
...
+    x_verts: np.ndarray = field(
+        metadata=schema(
+            description="The Nx1 x coordinates of the polygons vertices", min_len=4
+        )
+    )
+    y_verts: np.ndarray = field(
+        metadata=schema(
+            description="The Nx1 y coordinates of the polygons vertices", min_len=4
+        )
+    )
```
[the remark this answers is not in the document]
MichaelStubbings: Yes, I believe these get converted to ndarrays inside of matplotlib's patch functions. Have since changed this and they can be plotted without any problems

scanspec/regions.py
thomascobb: @EmmaCD what do you mean by regular polygons? Like a hexagon? If so, then I don't think anyone wants that. This is for "my sample is a funny shape, draw a polygon around it an only scan that bit"

scanspec/regions.py
```diff
@@ -259,26 +317,96 @@ class Circle(Region):
...
+    y_axis: str = field(
+        metadata=schema(description="The name matching the y axis of the spec")
+    )
+    x_middle: float = field(
+        metadata=schema(description="The central x point of the ellipse")
+    )
+    y_middle: float = field(
+        metadata=schema(description="The central y point of the ellipse")
+    )
+    x_semiaxis: float = field(
```
thomascobb, line 379: Hmm, the naming doesn't seem right, I've never heard of a semiaxis before, but we can't use width and height here as we are generic in x and y. How about we specify x_min, y_min, x_max, y_max like for a rectangle rather than the midpoints like the circle?

thomascobb (changes requested):

scanspec/regions.py
EmmaCD: Alternatively, x_semidiameter and y_semidiameter might be appropriate technical terms https://en.wikipedia.org/wiki/Semidiameter.

MichaelStubbings: https://en.wikipedia.org/wiki/Semi-major_and_semi-minor_axes

MichaelStubbings: Though realistically using radius is the most inclusive term

thomascobb: x_radius and y_radius?

PR #20, opened 2021-06-01
Smallest step

MichaelStubbings:
- Adds 'smallest step per axis' and 'absolute smallest step' parameters to the get_points query
- Absolute smallest step is handled as a resolver
- Removes optional types caused by dict.get commands

### DIFF
--- a/scanspec/service.py
+++ b/scanspec/service.py
@@ -17,39 +17,30 @@ from scanspec.specs import Spec
 class Points:
     """ A collection of singular or multidimensional locations in scan space"""
 
-    def __init__(self, points: Optional[np.ndarray]):
+    def __init__(self, points: np.ndarray):
         self._points = points
 
     @resolver
-    def string(self) -> Optional[str]:
+    def string(self) -> str:
         return np.array2string(self._points)
 
     @resolver
-    def float_list(self) -> Optional[List[float]]:
-        if self._points is None:
-            return None
-        else:
-            return self._points.tolist()
+    def float_list(self) -> List[float]:
+        return self._points.tolist()
 
     @resolver
-    def b64(self) -> Optional[str]:
-        if self._points is None:
-            return None
-        else:
-            # make sure the data is sent as float64
-            assert np.dtype(self._points[0]) == np.dtype(np.float64)
-            return base64.b64encode(self._points.tobytes()).decode("utf-8")
+    def b64(self) -> str:
+        # make sure the data is sent as float64
+        assert np.dtype(self._points[0]) == np.dtype(np.float64)
+        return base64.b64encode(self._points.tobytes()).decode("utf-8")
 
     # Self b64 decoder for testing purposes
     @resolver
-    def b64Decode(self) -> Optional[str]:
-        if self._points is None:
-            return None
-        else:
-            r = np.dtype(self._points[0])
-            s = base64.decodebytes(base64.b64encode(self._points.tobytes()))
-            t = np.frombuffer(s, dtype=r)
-            return np.array2string(t)
+    def b64Decode(self) -> str:
+        r = np.dtype(self._points[0])
+        s = base64.decodebytes(base64.b64encode(self._points.tobytes()))
+        t = np.frombuffer(s, dtype=r)
+        return np.array2string(t)
 
 
 @dataclass
@@ -62,11 +53,11 @@ class AxisFrames:
     """A fixed reference that can be scanned. i.e. a motor, time or
     number of repetitions.
     """
-    lower: Optional[Points]
+    lower: Points
     """The lower bounds of each midpoint (used when fly scanning)"""
-    midpoints: Optional[Points]
+    midpoints: Points
     """The centre points of the scan"""
-    upper: Optional[Points]
+    upper: Points
     """The upper bounds of each midpoint (used when fly scanning)"""
 
 
--- a/scanspec/service.py
+++ b/scanspec/service.py
@@ -92,7 +92,7 @@ class PointsResponse:
     """The number of frames returned by the getPoints query
     (controlled by the max_points argument)"""
     smallest_abs_step: SmallestStep
-    """The smallest step between midpoints across all axes in the scan"""
+    """The smallest step between midpoints across ALL axes in the scan"""
 
 
 # Checks that the spec will produce a valid scan
--- a/scanspec/service.py
+++ b/scanspec/service.py
@@ -31,8 +31,8 @@ class Points:
     @resolver
     def b64(self) -> str:
         # make sure the data is sent as float64
-        assert np.dtype(self._points[0]) == np.dtype(np.float64)
-        return base64.b64encode(self._points.tobytes()).decode("utf-8")
+        assert self._points.dtype == np.dtype(np.float64)
+        return base64.b64encode(self._points.tobytes()).decode()
 
     # Self b64 decoder for testing purposes
     @resolver
@@ -41,6 +41,9 @@ class Points:
         s = base64.decodebytes(base64.b64encode(self._points.tobytes()))
         t = np.frombuffer(s, dtype=r)
         return np.array2string(t)
+
+    def get_points(self) -> np.ndarray:
+        return self._points
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


scanspec/service.py
```diff
@@ -62,12 +62,30 @@ class AxisFrames:
...
+    smallest_step: float
+    """The smallest step between midpoints in this axis"""
+
+
+@dataclass
+class SmallestStep:
+    """ The smallest distance between midpoints in a multidimensional scan space"""
+
+    def __init__(self, points: List[ndarray]):
+        # points is an mxn array where n is the number of axes present in the scan
```
EmmaCD, line 80: I guess you mean it is an N-element list of M-element arrays?

scanspec/service.py
```diff
@@ -62,12 +62,30 @@ class AxisFrames:
...
     """The upper bounds of each midpoint (used when fly scanning)"""
+    smallest_step: float
+    """The smallest step between midpoints in this axis"""
+
+
+@dataclass
+class SmallestStep:
+    """ The smallest distance between midpoints in a multidimensional scan space"""
+
+    def __init__(self, points: List[ndarray]):
```
EmmaCD, line 79: I wonder if it makes sense to pass the whole `AxesPoints` dictionary here instead?

scanspec/service.py
```diff
@@ -7,7 +7,16 @@
 from aiohttp import web
 from apischema.graphql import graphql_schema, resolver
 from graphql_server.aiohttp.graphqlview import GraphQLView, _asyncify
-from numpy import array2string, dtype, float64, frombuffer, ndarray
+from numpy import (
```
thomascobb, line 10: The convention is `import numpy as np` then use them as `np.ndarray` etc.

scanspec/service.py
```diff
@@ -17,39 +26,30 @@
...
-    def b64(self) -> Optional[str]:
-        if self._points is None:
-            return None
-        else:
-            # make sure the data is sent as float64
-            assert dtype(self._points[0]) == dtype(float64)
-            return base64.b64encode(self._points.tobytes()).decode("utf-8")
+    def b64(self) -> str:
+        # make sure the data is sent as float64
+        assert dtype(self._points[0]) == dtype(float64)
```
thomascobb, line 43:
```suggestion
        assert self._points.dtype == np.float64
```

scanspec/service.py
```diff
@@ -17,39 +26,30 @@
...
+        return base64.b64encode(self._points.tobytes()).decode("utf-8")
```
thomascobb, line 44:
```suggestion
        return base64.b64encode(self._points).decode()
```

scanspec/service.py
```diff
@@ -17,39 +26,30 @@
...
     @resolver
-    def b64Decode(self) -> Optional[str]:
-        if self._points is None:
-            return None
-        else:
-            r = dtype(self._points[0])
-            s = base64.decodebytes(base64.b64encode(self._points.tobytes()))
-            t = frombuffer(s, dtype=r)
-            return array2string(t)
+    def b64Decode(self) -> str:
```
thomascobb, line 48: I'd put this in the unit tests rather than here

scanspec/service.py
```diff
@@ -62,12 +62,30 @@ class AxisFrames:
     """A fixed reference that can be scanned. i.e. a motor, time or
     number of repetitions.
     """
-    lower: Optional[Points]
+    lower: Points
     """The lower bounds of each midpoint (used when fly scanning)"""
```
thomascobb, line 66:
```suggestion
    """The lower bounds of each frame (used when fly scanning)"""
```

scanspec/service.py
```diff
@@ -62,12 +62,30 @@ class AxisFrames:
...
-    midpoints: Optional[Points]
-    """The centre points of the scan"""
-    upper: Optional[Points]
+    midpoints: Points
+    """The middle points of the scan"""
```
thomascobb, line 68:
```suggestion
    """The midpoints of each frame"""
```

scanspec/service.py
```diff
@@ -62,12 +62,30 @@ class AxisFrames:
...
+    upper: Points
     """The upper bounds of each midpoint (used when fly scanning)"""
```
thomascobb, line 70:
```suggestion
    """The upper bounds of each frame (used when fly scanning)"""
```

scanspec/service.py
thomascobb: I would argue this whole class should be a resolver on `PointsResponse`. This would make `PointsResponse` into a class like `Points` that was passed the whole chunk, then only calculated `AxisFrames` or `smallest_abs_step` on demand. `AxisFrames` would probably look similar...

scanspec/service.py
```diff
@@ -104,10 +128,13 @@ def get_points(spec: Spec, max_frames: Optional[int] = 200000) -> PointsResponse
...
+
+    # TOTAL FRAMES
     total_frames = len(path)  # Capture the total length of the path
 
-    # Limit the consumed data to the max_frames argument
-    # # WARNING: path object is consumed after this statement
+    # MAX FRAMES | RETURNED FRAMES
+    # Limit the consumed data by the max_frames argument
+    # WARNING: path object is consumed after this statement
     if max_frames is None:
```
thomascobb, line 138: No matter what the value of max_frames, we should be able to write:
```python
chunk = path.consume(max_frames)
returned_frames = len(chunk)
```
If we can't then path.consume is wrong...

thomascobb (changes requested):

scanspec/service.py
MichaelStubbings: Tested in graphiql and seems to work as intended

scanspec/service.py
thomascobb: But isn't `len(points) == m  # number of axes present in the scan (rows)`
and `len(points[i]) == n  # n is the number of points present in each axis`?

So for `Line(y, 0, 1, 10) * Line(x, 0, 1, 5)` then m is 2 and n is 50

scanspec/service.py
EmmaCD: @MichaelStubbings I vaguely recall us talking about this `if max_frames is None` bit in the context of mypy complaining. Do you remember what the problem was?

scanspec/service.py
MichaelStubbings: .tobytes() is needed to stop mypy complaining: Argument 1 to "b64encode" has incompatible type "ndarray"; expected "bytes"

scanspec/service.py
MichaelStubbings: Oh I understand. Yes, for some reason I had it in my head that it was a list with a single mxn array in it. 

So the list dimension is the columns, (m) being the number of axes present
Then the arrays within the list, (n) are the number of points present in each axis
As you mentioned Tom

scanspec/service.py
thomascobb: Fair enough, performance doesn't seem affected much so that's fine:
```
$ python -m timeit -s "import numpy as np; import base64; a = np.arange(100000)" "base64.b64encode(a.tobytes())"
200 loops, best of 5: 1.69 msec per loop
$ python -m timeit -s "import numpy as np; import base64; a = np.arange(100000)" "base64.b64encode(a)"
200 loops, best of 5: 1.68 msec per loop
```
If performance was affected we could add `# type: ignore` to it

scanspec/service.py
MichaelStubbings: This was a great suggestion. Not only should it be much more performant, but it also reduces the number of fields you need to request in graphql.

scanspec/service.py
```diff
@@ -76,12 +69,60 @@ class PointsResponse:
...
-    axes: List[AxisFrames]
     total_frames: int
     returned_frames: int
 
+    def __init__(
+        self, chunk: Dimension, total_frames: int, returned_frames: int,
+    ):
+        self.total_frames = total_frames
+        """The number of frames present across the entire spec"""
+        self.returned_frames = returned_frames
```
thomascobb, line 80:
```suggestion
        self.returned_frames = len(chunk)
```

scanspec/service.py
```diff
@@ -62,12 +53,14 @@ class AxisFrames:
...
-    """The centre points of the scan"""
-    upper: Optional[Points]
-    """The upper bounds of each midpoint (used when fly scanning)"""
+    lower: Points
+    """The lower bounds of each frame (used when fly scanning)"""
+    midpoints: Points
+    """The midpoints of each frame"""
+    upper: Points
+    """The upper bounds of each frame (used when fly scanning)"""
+    smallest_step: float
```
thomascobb, line 62: Can we make this a resolver too?

thomascobb (changes requested):
Almost there, looking much cleaner now

scanspec/service.py
MichaelStubbings: returned frames? if so then we can. 

If we wanted total frames as well then we would have to pass the path object before it has been affected by the max_points

thomascobb: Sorry, I meant for `smallest_step`, that's the only thing that takes time to calculate

scanspec/service.py
thomascobb: Then you don't have to pass in returned_frames either

scanspec/service.py
```diff
@@ -104,38 +142,26 @@ def get_points(spec: Spec, max_frames: Optional[int] = 200000) -> PointsResponse
...
-    scan_points = [
-        AxisFrames(
-            axis,
-            Points(chunk.lower.get(axis)),
-            Points(chunk.midpoints.get(axis)),
-            Points(chunk.upper.get(axis)),
-        )
-        for axis in spec.axes()
-    ]
+        chunk = path.consume(max_frames)
```
thomascobb, line 162: The whole of the if statement can now be:
```python
    if max_frames and max_frames > len(path):
        # Cap the frames by the max limit
        path = reduce_frames(dims, max_frames)
    chunk = path.consume()
```

PR #24, opened 2021-07-05
Add helper factory method to Static spec to create a scan with no motion, rename TIME dimension to DURATION

callumforrester:
The `Static` spec can be used to create a scan that involves no axes except for `TIME` via `Static(TIME, duration, num)`, this translates to "do nothing for `duration` seconds `num` times". This PR adds a helper method to do this without having to refer to `TIME`, it can be used like so:

```python
spec = Static.duration(duration, num)
```

This PR also renames `TIME` to `DURATION` since `TIME` implies absolute time when the axis is used to refer to the relative time from the start of a particular point.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -65,10 +65,10 @@
         },
         "apischema": {
             "hashes": [
-                "sha256:01b6dab6bebd2ea1c618c3831a3723ae8b1b38139816cb9c1a8f75d7b46c9047",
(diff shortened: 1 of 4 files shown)


thomascobb (approved):
Did you have an idea of how to fix the CI failures?

callumforrester:
The CI failures are caused by a mypy error that is also present on master, it can be solved by installing the type stubs:
```bash
mypy --install-types
```

Not sure how to make Pipenv do it though

PR #25, opened 2021-07-06
Update docs and CI

thomascobb:


thomascobb:
mypy errors will go away when sphinx_graphql exists

PR #33, opened 2021-08-17
Add Dimension.gap

thomascobb:


docs/how-to/iterate-a-spec.rst
```diff
@@ -87,3 +87,30 @@ You can also use this method to only run a subset of the scan:
...
+>>> from scanspec.specs import Line, fly
+>>> grid = fly(Line("y", 0, 1, 2) * ~Line("x", 1, 2, 3), 0.1)
+>>> chunk = grid.path().consume()
+>>> chunk.midpoints["x"]
+array([1. , 1.5, 2. , 2. , 1.5, 1. ])
+
+You can check where the gaps are by using the `Dimension.gap` attribute:
+
+>>> chunk.gap
+array([ True, False, False,  True, False, False])
```
callumforrester, line 107: I can see the concrete use of this, but in what is a gap in abstract terms? Is it just an arbitrary way of grouping the points? If so, rather than splitting the points into "gapped" and "not gapped" you could split them into groups. For example, in this case:

```python
>>> chunk.midpoints["x"]
array([1. , 1.5, 2. , 2. , 1.5, 1. ])

>>> chunk.groups
array([0, 0, 0, 1, 1, 1])
```
So gapped points are not special. Unsure if this is cleaner or not.

scanspec/core.py
```diff
@@ -176,23 +176,31 @@ def if_instance_do(x, cls: Type, func: Callable):
         return NotImplemented
 
 
+def last_first_gap(upper: AxesPoints, lower: AxesPoints) -> bool:
```
callumforrester, line 179: Not sure about this function name for readability purposes. How about `is_gap_between`?

scanspec/specs.py
```diff
@@ -303,6 +344,9 @@ class Concat(Spec):
             description="The right-hand Spec to Concat, midpoints will appear later"
         ),
     ]
+    gap: A[
+        bool, schema(description="If True, force a gap in the output at the join")
```
callumforrester, line 348: I can see many specs needs to deal with the gapped and non-gapped cases explicitly like this. Is there a way we can make it part of the composition instead? I am imagining something like

```python
Concat(spec_a, Concat(GAP, spec_b))

# ...or

Concat(spec_a, gapped(spec_b))
```

scanspec/core.py
thomascobb: Maybe `is_gap_between(dim1, dim2)`

scanspec/specs.py
thomascobb: I think it might have to be

docs/how-to/iterate-a-spec.rst
thomascobb: I think from a processing point of view the existing structure is more useful

PR #34, opened 2021-09-29
Make Spec, Frames and AxesPoints generic

callumforrester:
- Make `AxesPoints` key a generic type rather than a string
- Propagate genericness through `Frames`, `Spec` and all subclasses of `Spec` such that you can now have, for example, `spec: Spec[str]` or `spec: Spec[Motor]`
- Set service code to explicitly use `Spec[str]`
- Fix unit test debug configuration


scanspec/regions.py
```diff
@@ -39,7 +42,7 @@ def axis_sets(self) -> List[Set[str]]:
         this region spans"""
         raise NotImplementedError(self)
 
-    def mask(self, points: AxesPoints) -> np.ndarray:
+    def mask(self, points: AxesPoints[K]) -> np.ndarray:
```
thomascobb, line 45: Do we need to make Region a generic too?

callumforrester: I think so, I just pushed it, let me know if I missed anywhere.

callumforrester:
Currently getting issues with generic union, see: https://github.com/wyfo/apischema/issues/207

callumforrester:
@thomascobb I think this is nearly there but it may be worth a rethink. Mypy wants type annotations for specs much more often (and in more awkward places) than I had hoped. 

One reason for this is that it's not good at retroactively inferring types. So if I do
```python
empty = Static.duration(0.5) * Line("x", 0, 1, 2)
```
It will demand a type annotation because it cannot infer that the axes for the `Static` part of the expression are of type `str` from the fact that I passed a string to the `Line`.

Another problem is the fact that `TypeVar`s need to be bound inside a class.
For example, for line:
```python
AAxis = A[K, schema(description="An identifier for what to move")]


@dataclass
class Line(Spec[K]):
    ...
```

Putting `K` in `AAxis` isn't good enough, the reference to `K` needs to be inside `Line` for mypy to bind it. Otherwise you have to type such things as 
```python
# I'm passing a string so mypy should know that K=str, but it doesn't because the reference to K is outside Line
spec: Line[str] = Line("x", 0, 10, 5)
```
We can fix this by moving `AAxis` into `Line` but that requires some duplication and I'm not sure how often this pattern is used in the rest of the library.

What are your thoughts?

callumforrester:
The second problem was fixed with
```python
class Line:
    axis: AAxis[K]
```

PR #41, opened 2021-11-25
Test/gh pages

BenBradnick:
Testing creating a version file for the gh-pages branch based on the available builds.

### DIFF
--- a/docs/github_pages_version_file.py
+++ b/docs/github_pages_version_file.py
@@ -85,7 +85,9 @@ def push_file_to_GitHub(file_string: str, filename: str, remote: str, branch: st
         # Status of zero means tracked. So check if file has changed
         if not status:
             status = run_command(
-                f"git diff --exit-code {filename}", return_exit_status=True
+                f"git diff --exit-code {filename}",
+                return_exit_status=True,
+                fail_on_bad_status=False,
             )
 
         # If status is one (either untracked or file has changed) then commit the file
--- a/docs/github_pages_version_file.py
+++ b/docs/github_pages_version_file.py
@@ -44,27 +44,6 @@ def get_branch_contents(remote: str, branch: str) -> List[str]:
     return stdout.strip().split("\n")
 
 
-def get_docs_url(github_org: str, project: str, version: str):
-    """Get the URL for a specific build of the project."""
-    return f"https://{github_org}.github.io/{project}/{version}"
-
-
-def get_version_file_string(
-    main_branches: List[str], branches: List[str], releases: List[str],
-) -> str:
-    """Create the string for the RST file."""
-    file_str = ""
-    # Add branches first
-    for version in main_branches:
-        file_str += f"{version}\n"
-    for version in branches:
-        file_str += f"{version}\n"
-    # Add releases
-    for version in releases:
-        file_str += f"{version}\n"
-    return file_str
-
-
 def push_file_to_GitHub(file_string: str, filename: str, remote: str, branch: str):
     """Push the version file to the GitHub Pages branch."""
     # Set git config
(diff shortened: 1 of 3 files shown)


docs/github_pages_version_file.py
```diff
@@ -0,0 +1,175 @@
...
+    tags = get_sorted_tags_list()
+
+    # Get the directories (i.e. builds) from the GitHub Pages branch
+    all_build_versions = get_branch_contents(remote, github_pages_branch)
+
+    # Parse all builds, storing releases in an unordered dict
+    unsorted_releases: Dict[str, GitHubPagesBuild] = {}
+    for version in all_build_versions:
+        # Create our instance
+        url = get_docs_url(github_org, project, version)
```
thomascobb, line 151: Don't need the URL, just the name

docs/github_pages_version_file.py
```diff
@@ -0,0 +1,175 @@
...
+        if status:
+            print(f"File {filename} being commited to {remote}/{branch}.")
+            # Add the file
+            run_command(f"git add {filename}")
+
+            # Commit the file
+            run_command("git commit -qm Update-version-file")
+
+            # Push to remote
+            run_command("git push -q")
```
thomascobb, line 114: We have permissions to do this? That makes things easier!

thomascobb: Could we do the deploy of the build/html directory too then? Save our use of the external action:
https://github.com/dls-controls/scanspec/blob/test/gh-pages/.github/workflows/docs.yml#L41-L60

thomascobb (commented):
I think we can cut this down a bit, then we could make it a script in https://github.com/dls-controls/sphinx_rtd_theme_github_versions

docs/github_pages_version_file.py
BenBradnick: Now just creates a text file of a list of versions: https://github.com/dls-controls/scanspec/blob/test/gh-pages-file/version.txt

docs/github_pages_version_file.py
BenBradnick: Looks like it! I just ran this build which commits the file: https://github.com/dls-controls/scanspec/runs/4379379099?check_suite_focus=true

The file was committed to the test branch here: https://github.com/dls-controls/scanspec/blob/test/gh-pages-file/version.txt

BenBradnick:
> I think we can cut this down a bit, then we could make it a script in https://github.com/dls-controls/sphinx_rtd_theme_github_versions

How should this be cut down and how does the script end up being called from the sphinx repo?

docs/github_pages_version_file.py
```diff
@@ -0,0 +1,164 @@
...
+    stdout = run_command(
+        f"git ls-tree -d --name-only {remote}/{branch}", return_stdout=True
+    )
+    assert isinstance(
+        stdout, str
+    ), f"Error getting directory contents of {remote}/{branch}"
+    return stdout.strip().split("\n")
+
+
+def get_docs_url(github_org: str, project: str, version: str):
```
thomascobb, line 47: This function is unused?

docs/github_pages_version_file.py
```diff
@@ -0,0 +1,164 @@
...
+        else:
+            branches.append(version)
+
+    # Use output from git tag to sort the releases
+    for tag in tags:
+        if tag in unsorted_releases:
+            releases.append(tag)
+
+    # Get version file string
+    version_file_string = get_version_file_string(main_branches, branches, releases)
```
thomascobb, line 155:
```suggestion
    version_file_string = "\n".join(main_branches + branches + releases)
```

docs/github_pages_version_file.py
```diff
@@ -0,0 +1,164 @@
...
+    ), f"Error getting directory contents of {remote}/{branch}"
+    return stdout.strip().split("\n")
+
+
+def get_docs_url(github_org: str, project: str, version: str):
+    """Get the URL for a specific build of the project."""
+    return f"https://{github_org}.github.io/{project}/{version}"
+
+
+def get_version_file_string(
```
thomascobb, line 52: ...which would make this function unused

docs/github_pages_version_file.py
BenBradnick: I always forget about join...

thomascobb:
Code moved to https://github.com/dls-controls/sphinx_rtd_theme_github_versions

PR #44, opened 2021-12-06
Add pre-commit and use it in CI

thomascobb:


.github/setup_action/action.yml
```diff
@@ -0,0 +1,42 @@
...
+    required: false
+    description: What command to pass to pipenv run
+    default: ""
+  python:
+    required: false
+    description: Version of Python to install
+    default: "3.7"
+  pipenv:
+    required: false
+    description: Either skip-lock or deploy
```
BenBradnick, line 15: I'm guessing there's not a nicer way to have this as a bool/choice type?

thomascobb:
Closed in favour of #46

PR #45, opened 2022-01-07
Remove duplicate draw logic

thomascobb:


thomascobb:
`skip-lock` builds fail because of https://github.com/wyfo/apischema/issues/316

PR #48, opened 2022-03-14
Remove restrictions on axis order for concatenation

callumforrester:
Allow concatenating `Frames` and `Spec` obejcts with axes in different orders, so long as the set of axes is the same. Order is inherited from the lefthand side.

```python
# Valid, no change
Concat(Line("x", 0, 10, 10), Line("x", 10, 20, 10))

# Valid, no change
Concat(Line("x", 0, 10, 10) * Line("y", 0, 10, 10), Line("x", 10, 20, 10) * Line("y", 0, 20, 10))

# Valid, axis order is ["x", "y"]
Concat(Line("y", 0, 10, 10) * Line("x", 0, 10, 10), Line("x", 10, 20, 10) * Line("y", 0, 20, 10))

# Invalid, axes are not the same
Concat(Line("y", 0, 10, 10) * Line("x", 0, 10, 10) * Line("z", 0, 10, 10), Line("x", 10, 20, 10) * Line("y", 0, 20, 10))
```

We currently consider this reasonable because nothing downstream depends on knowing the order, so it is reasonable to abstract the information away in exchange for flexibility. Unsure if this will change though.

### DIFF
--- a/src/scanspec/core.py
+++ b/src/scanspec/core.py
@@ -299,18 +299,21 @@ class Frames(Generic[Axis]):
     def concat(self, other: Frames[Axis], gap: bool = False) -> Frames[Axis]:
         """Return a new Frames object concatenating self and other.
 
-        Requires both Frames objects to have the same axes.
+        Requires both Frames objects to have the same axes, but not necessarily in
+        the same order. The order is inherited from self, so other may be reordered.
 
         Args:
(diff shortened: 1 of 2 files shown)


coretl:
You seem to have a newer version of black, so linting is failing. In setup.cfg `black==19.10b0` is specified, please could you reformat with that?

PR #50, opened 2022-04-28
Update to use setuptools_scm instead of versiongit

GDYendell:


GDYendell:
Needs rebasing on master when #49 goes int

src/scanspec/sphinxext.py
```diff
@@ -2,7 +2,7 @@
 
 from matplotlib.sphinxext import plot_directive
 
-from ._version_git import __version__
+from ._version import __version__
```
coretl, line 5:
```suggestion
from scanspec import __version__
```

pyproject.toml
```diff
@@ -2,5 +2,8 @@
 # To get a reproducible wheel, wheel must be pinned to the same version as in
 # dls-python3, and setuptools must produce the same dist-info. Cap setuptools
 # to the last version that didn't add License-File to METADATA
-requires = ["setuptools<57", "wheel==0.33.1"]
+requires = ["setuptools>=45,<57", "setuptools_scm[toml]>=6.2", "wheel==0.33.1"]
 build-backend = "setuptools.build_meta"
+
+[tool.setuptools_scm]
+write_to = "src/scanspec/_version.py"
```
coretl, line 9: This is also in setup.py, can we omit one of them?

setup.py
```diff
@@ -1,13 +1,9 @@
...
-# Import <package>._version_git.py without importing <package>
-path = glob.glob(__file__.replace("setup.py", "src/*/_version_git.py"))[0]
-spec = importlib.util.spec_from_file_location("_version_git", path)
-vg = importlib.util.module_from_spec(spec)
-spec.loader.exec_module(vg)
-
-setup(cmdclass=vg.get_cmdclass(), version=vg.__version__)
+setup(
+    use_scm_version={
+        "write_to": "src/scanspec/_version.py"
```
coretl, line 6: This is also in pyproject.toml, can we omit one of them?

setup.py
```diff
@@ -1,13 +1,9 @@
...
+    },
+    setup_requires=["setuptools>=45,<57", "setuptools_scm[toml]>=6.2", "wheel==0.33.1"],
```
coretl, line 8: Can all of this go in setup.cfg instead?

pyproject.toml
coretl: Or can it go in setup.cfg?

setup.py
coretl: Actually, can it just be reduced to setuptools_scm[toml]>=6.2?

coretl: In setup.cfg...

.gitignore
```diff
@@ -22,6 +22,8 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+src/scanspec/_version.py
```
coretl, line 25:
```suggestion
src/*/_version.py
```

pyproject.toml
coretl: Is it still required if we put things in setup.cfg?

setup.py
coretl: I mean still leave it as you have it for pyproject.toml, but delete the wheel and setuptools references in setup.cfg?

GDYendell: I am not sure that we need a pyproject.toml at all, but pipenv makes one if it doesn't exist with this content:
```
[build-system]
requires = ["setuptools", "wheel"]
build-backend = "setuptools.build_meta"
```

pyproject.toml
GDYendell: No `python -m setuptools_scm` won't use the config from setup.cfg. But, it doesn't really matter unless we want to check the version parsing works with any additional config, such as as tag_regex to allow a `+dlsX` suffix. Even then, you can do `python setup.py --version` instead.

setup.py
GDYendell: Taking out the pinned wheel version has fixed [this](https://github.com/dls-controls/scanspec/runs/6214158036?check_suite_focus=true#step:3:105) error

GDYendell: We have to have the setup_requires in setup.py or setup.cfg for Pipenv to work, so I suggest we have it in setup.cfg and have the minimal setup.py and pyproject.toml

pyproject.toml
```diff
@@ -1,6 +1,3 @@
 [build-system]
-# To get a reproducible wheel, wheel must be pinned to the same version as in
-# dls-python3, and setuptools must produce the same dist-info. Cap setuptools
-# to the last version that didn't add License-File to METADATA
-requires = ["setuptools<57", "wheel==0.33.1"]
+requires = ["setuptools", "wheel"]
```
coretl, line 2: Can we put this back please for back compat

setup.cfg
```diff
@@ -21,12 +21,16 @@ packages = find:
...
+setup_requires =
+    setuptools>=45
+    setuptools_scm[toml]>=6.2
 install_requires =
     # make sure a python 3.9 compatible numpy is selected
     numpy>=1.19.3
     click
     apischema>=0.17.2
     typing_extensions
+    setuptools_scm[toml]>=6.2
```
coretl, line 33: This is only required at dev time, so can you move this to the dev extra please?

setup.cfg
```diff
@@ -102,8 +106,8 @@ testpaths =
     docs src tests
 
 [coverage:run]
-# This is covered in the versiongit test suite so exclude it here
-omit = */_version_git.py
+# Ignore generated _version.py
+omit = src/*/_version.py
```
coretl, line 110: If we delete this line do we still get the same coverage result? If so, it can go

src/scanspec/sphinxext.py
```diff
@@ -2,7 +2,7 @@
 
 from matplotlib.sphinxext import plot_directive
 
-from ._version_git import __version__
+from ._version import version as __version__
```
coretl, line 5:
```suggestion
from scanspec import __version__
```

setup.cfg
GDYendell: It does not stay the same. We get 2 more statements, but they are covered.

coretl: That sounds fine to me

pyproject.toml
GDYendell: Pinning wheel seems to make `build` unhappy. https://github.com/dls-controls/scanspec/runs/6226284583?check_suite_focus=true#step:3:107

coretl: because you also have it in `setup_requires`. I suggest you have:
pyproject.toml: `requires = ["setuptools<57", "wheel==0.33.1"]`
setup.cfg: `setup_requires: setuptools_scm[toml]>=6.2` and the same in the dev dependency

GDYendell: OK I think I have now basically gone in a circle. Is [7c58fc3](https://github.com/dls-controls/scanspec/pull/50/commits/7c58fc39cd72cc469c43bb812a8d130a0ca05a06) functionally different to what I had initially in [198f409](https://github.com/dls-controls/scanspec/pull/50/commits/198f4092b1e3830fad8af86dcae45de1a8069f38)?

coretl: Pretty much the same. There's an interesting POV here: https://github.com/pypa/setuptools/pull/2824#issuecomment-968464092

coretl: Does `dls-python3 setup.py bdist_wheel` work with what you have now?

GDYendell:
I am not sure why this change adds wheel to Pipfile.lock or if it should.

coretl:
Superceded by #49

Issue #52, opened 2022-10-19
Use pydantic instead of apischema


callumforrester:
As discussed

PR #56, opened 2022-10-28
Convert Service to REST using FastAPI and Make ScanSpecs Pydantic Models

callumforrester:
As discussed offline, merging #52 and #53 into one big PR.
 
* Replace the GraphQL service with a FastAPI REST service and 
* Convert ScanSpecs to pydantic models.
* Add a new decorator `@discriminated_union_of_subclasses` to enable similar tagged union behaviour to the GraphQL service
* Write tests for REST service
* Write API documentation
* Remove all references to GraphQL


src/scanspec/specs.py
```diff
@@ -619,17 +596,18 @@ def calculate(self, bounds=True, nested=False) -> List[Frames[Axis]]:
             self._spiral_from_indexes, self.axes(), self.num, bounds
         )
 
-    @alternative_constructor
```
callumforrester, line 622: Proxy thread for discussion in previous PR: https://github.com/dls-controls/scanspec/pull/54#discussion_r1007061793

callumforrester: I'm now tempted to say that we don't need alternative constructors for the REST interface, since REST is all about specifying canonical data models. In other words, the model of a `Line` is `axis`, `start`, `stop` and `num`, and if DAQ want to specify it a different way then the conversion is our responsibility client-side. This potentially creates a little duplication but also ensures separation of concerns. Happy to discuss though. If we do loose alternative constructors, we may also wish to ask whether the `/valid` endpoint is even needed any more.

callumforrester: From discussion: Alternative constructors may be needed, will take the agile approach and reintroduce them when we make API clients. The behaviour is still available in native Python via `classmethod`s.

src/scanspec/core.py
```diff
@@ -37,140 +31,159 @@
...
+
+
+def _discriminated_union_of_subclasses(
+    super_cls: Type,
+    discriminator: str,
+    config: Optional[Type[BaseConfig]] = None,
+) -> Union[Type, Callable[[Type], Type]]:
+
+    super_cls.__ref_classes__ = set()
+    super_cls.__model__ = None
```
coretl, line 139: Python stdlib reserves the right to use any dunder methods for its own methods. Suggest we keep a couple of global dicts around for this instead of attaching to the cls

src/scanspec/service.py
```diff
@@ -1,161 +1,302 @@
...
+            An error otherwise.
     """
-    # [array[1] - array[0], array[2] - array[1], ...]
-    adjacent_diffs = array[1:] - array[:-1]
-    return np.absolute(adjacent_diffs)
+    valid_spec = Spec.deserialize(spec.serialize())
+    return ValidResponse(spec, valid_spec)
+
+
+@app.post("/midpoints", response_model=MidpointsResponse)
```
coretl, line 142: What does GDA actually want from this service? If just midpoints then this is fine, if midpoints, bounds, smallest step, then we calculate the spec 3 times which is quite wasteful

src/scanspec/service.py
```diff
@@ -186,40 +327,58 @@ def sub_sample(frames: Frames[str], ratio: float) -> Frames:
...
-        cors_config = aiohttp_cors.setup(
-            app,
-            defaults={
-                "*": aiohttp_cors.ResourceOptions(
-                    allow_credentials=True,
-                    expose_headers="*",
-                    allow_headers="*",
-                )
-            },
+        app.add_middleware(
```
coretl, line 357: Do we need CORS now? I only added it for graphiql...

src/scanspec/service.py
```diff
@@ -186,40 +327,58 @@ def sub_sample(frames: Frames[str], ratio: float) -> Frames:
...
-        # Configure CORS on all routes.
-        for route in list(app.router.routes()):
-            cors_config.add(route)
+    import uvicorn
+
+    uvicorn.run(app, port=port)
+
 
-    web.run_app(app, port=port)
+def scanspec_schema_text() -> str:
```
coretl, line 370: What do we use the schema.json for? Is it for upload to api portal?

src/scanspec/specs.py
```diff
@@ -582,16 +560,16 @@ class Spiral(Spec[Axis]):
...
-    rotate: A[
-        float, schema(description="How much to rotate the angle of the spiral")
-    ] = 0.0
+    x_axis: Axis = Field(description="An identifier for what to move for x")
+    y_axis: Axis = Field(description="An identifier for what to move for y")
+    x_start: float = Field(description="x centre of the spiral")
+    y_start: float = Field(description="y centre of the spiral")
+    x_range: float = Field(description="x width of the spiral")
+    y_range: float = Field(description="y width of the spiral")
+    num: int = Field(min=1, description="Number of frames to produce")
```
coretl, line 569: Can we share Field defs in multiple classes?

src/scanspec/core.py
```diff
@@ -37,140 +31,159 @@
...
+    "discriminated_union_of_subclasses",
+    "StrictConfig",
 ]
 
 
-def _rec_subclasses(cls: type) -> Iterator[type]:
-    for sub_cls in cls.__subclasses__():
-        yield sub_cls
-        yield from _rec_subclasses(sub_cls)
+class StrictConfig(BaseConfig):
```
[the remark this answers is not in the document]
callumforrester: I can't think of any downsides to forbidding it, and it is closer to the previous behaviour. I think it is probably best if the definition of each scanspec is fairly formal and a user is notified early on if it is "wrong", rather than thinking there is an extra argument in `Line` for 6 months before realising there is a problem.

src/scanspec/service.py
callumforrester: I suspect just midpoints initially, bounds and smallest step may come later. I was going to suggest adding a caching mechanism when required, since I doubt it will want the extra information for every request.

src/scanspec/specs.py
callumforrester: Yes, but it saves significantly less code than before because you can't specify different defaults, before you had
```python
ANum = A[int, schema(min=1, description="Number of frames to produce")]

class Foo:
    num: ANum

class Bar:
    num: ANum = 1
```

Whereas now we need
```python
NumFrames = Field(min=1, description="Number of frames to produce")
NumFramesDefault1 = Field(min=1, description="Number of frames to produce", default=1)

class Foo:
    num: NumFrames

class Bar:
    num: NumFramesDefault1
```

Because the default is defined in the same place as everything else.

src/scanspec/service.py
coretl: I guess that is the REST way, caching and multiple requests, rather than multiple arguments changing what comes out in the response

src/scanspec/core.py
coretl: It looks like you only use these in `__validate__`, so they could live anywhere if this is the case

src/scanspec/specs.py
coretl: Doesn't pydantic support `Annotated[int, Field(...)]`?

callumforrester: Don't think so:
```shell
Traceback (most recent call last):
  File "/home/vid18871/projects/scanspec/tests/test_specs.py", line 7, in <module>
    from scanspec.specs import (
  File "/home/vid18871/projects/scanspec/src/scanspec/specs.py", line 140, in <module>
    class Repeat(Spec[Axis]):
  File "pydantic/dataclasses.py", line 224, in pydantic.dataclasses.dataclass.wrap
  File "pydantic/dataclasses.py", line 342, in pydantic.dataclasses._add_pydantic_validation_attributes
  File "pydantic/dataclasses.py", line 397, in pydantic.dataclasses.create_pydantic_model_from_dataclass
  File "pydantic/main.py", line 1027, in pydantic.main.create_model
  File "pydantic/main.py", line 198, in pydantic.main.ModelMetaclass.__new__
  File "pydantic/fields.py", line 497, in pydantic.fields.ModelField.infer
  File "pydantic/fields.py", line 476, in pydantic.fields.ModelField._get_field_info
ValueError: cannot specify `Annotated` and value `Field`s together for 'num'
```

src/scanspec/core.py
callumforrester: Good point, that works, although I'm unsure if using globals where we don't absolutely need to is any better than using dunders. It would be far better if this information were attached to the class it was about. Can we just turn it into `super_cls._models`?

src/scanspec/service.py
callumforrester: https://github.com/dls-controls/scanspec/issues/78

src/scanspec/specs.py
callumforrester: Caused by https://github.com/pydantic/pydantic/issues/3496
Have added TODO for when it's fixed

src/scanspec/core.py
callumforrester: Have renamed to `super_cls._ref_classes` and `super_cls._models`

PR #81, opened 2023-02-07
Add shape utility method to Spec to get a scan's simplified shape

callumforrester:
* Add `Spec.shape()` method which gives the the number of dimensions in the scan in the simplest possible view
* Add tests for shape method


callumforrester:
Issue for mypy problems: https://github.com/dls-controls/scanspec/issues/82

coretl (approved):

PR #93, opened 2024-01-29
Move function for getting frame duration from dls-bluesky-core (from i22-bluesky) to ScanSpec

DiamondJoseph:
This function was added during i22 plan development, but appears to not be used in any current i22 plans.

If it is generally useful behaviour for a ScanSpec, it should probably exist in this repository rather than in dodal.

### DIFF
--- a/src/scanspec/specs.py
+++ b/src/scanspec/specs.py
@@ -667,3 +667,31 @@ def step(spec: Spec[Axis], duration: float, num: int = 1) -> Spec[Axis]:
         spec = step(Line("x", 1, 2, 3), 0.1)
     """
     return spec * Static.duration(duration, num)
+
+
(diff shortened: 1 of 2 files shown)


src/scanspec/specs.py
```diff
@@ -667,3 +667,31 @@ def step(spec: Spec[Axis], duration: float, num: int = 1) -> Spec[Axis]:
         spec = step(Line("x", 1, 2, 3), 0.1)
     """
     return spec * Static.duration(duration, num)
+
+
+def duration(frames: List[Frames]) -> Optional[float]:
```
coretl, line 672: Actually, should this be called something like `get_constant_duration`?

DiamondJoseph: I had renamed the function, but I've put it back to get_constant_duration.

PR #127, opened 2024-08-06
move to pydantic v2

### LINKED ISSUE #122: Move to pydantic 2

ZohebShaikh:
closes #122


pyproject.toml
```diff
@@ -80,6 +80,7 @@ ignore_missing_imports = true # Ignore missing stubs in imported modules
 # Run pytest with all our checkers, and don't spam us with massive tracebacks on error
 addopts = """
     --tb=native -vv --doctest-modules --doctest-glob="*.rst"
+    -W ignore::DeprecationWarning
```
DiamondJoseph, line 83: Make sure to get rid of this

src/scanspec/regions.py
```diff
@@ -93,87 +95,6 @@ def _merge_axis_sets(axis_sets: List[Set[Axis]]) -> Iterator[Set[Axis]]:
             yield axis_set
 
 
-@dataclass(config=StrictConfig)
```
DiamondJoseph, line 96: Re-order these back to how the order was, now that we've made the changes necessary for the Spec referencing types can go at the top still

src/scanspec/core.py
```diff
@@ -36,24 +46,15 @@
...
+StrictConfig: ConfigDict = {"extra": "forbid"}
 
 
-def discriminated_union_of_subclasses(
-    super_cls: Optional[Type] = None,
-    *,
-    discriminator: str = "type",
-    config: Optional[Type[BaseConfig]] = None,
-) -> Union[Type, Callable[[Type], Type]]:
+def discriminated_union_of_subclasses(cls):
```
DiamondJoseph, line 52: Let's leave discriminator as a possible override if possible, it just needs to be passed into _TaggedUnion's __init__ and then used on line 168

src/scanspec/core.py
```diff
@@ -117,72 +118,120 @@ def calculate(self) -> int:
...
+
+
+T = TypeVar("T", type, Callable)
+
+
+def deserialize_as(cls, obj):
+    return TypeAdapter(_tagged_unions[cls]._make_union()).validate_python(obj)
+
+
+def uses_tagged_union(cls_or_func: T) -> T:
```
DiamondJoseph, line 137: Document how/why this is required

src/scanspec/service.py
```diff
@@ -123,8 +127,9 @@ class SmallestStepResponse:
 
 
 @app.post("/valid", response_model=ValidResponse)
+@uses_tagged_union
 def valid(
-    spec: Spec = Body(..., examples=[_EXAMPLE_SPEC]),
+    spec: Spec = Body(..., examples=[_EXAMPLE_SPEC], discriminator="type"),
```
DiamondJoseph, line 132: Was `, discriminator="type"` required? Is it still with the other changes?

src/scanspec/specs.py
```diff
@@ -108,7 +120,256 @@ def serialize(self) -> Mapping[str, Any]:
     @classmethod
     def deserialize(cls, obj):
         """Deserialize the spec from a dictionary."""
-        return parse_obj_as(cls, obj)
+        return deserialize_as(cls, obj)
```
DiamondJoseph, line 123: re-order to how it was as above

src/scanspec/specs.py
```diff
@@ -108,7 +120,256 @@ def serialize(self) -> Mapping[str, Any]:
...
+        if num == 1:
+            # One point, stop will only be used for step size
+            stop = upper + half_step
+        else:
+            # Many points, stop will be produced
+            stop = upper - half_step
+        return cls(axis, start, stop, num)
+
+
+Line.bounded = validate_call(Line.bounded)
```
DiamondJoseph, line 236: Document why this is necessary (field validation for class methods that return a forward referenced: forward reference not available when the decorator would be processed)

DiamondJoseph (changes requested):

src/scanspec/service.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
```
DiamondJoseph, line 2:
```suggestion
```

Isn't being used anyway

src/scanspec/specs.py
```diff
@@ -506,6 +510,9 @@ def bounded(
         return cls(axis, start, stop, num)
 
 
+Line.bounded = validate_call(Line.bounded)  # type:ignore
```
DiamondJoseph, line 513: Document why this was required

src/scanspec/specs.py
```diff
@@ -106,10 +110,57 @@ def serialize(self) -> Mapping[str, Any]:
...
-    @classmethod
-    def deserialize(cls, obj):
+    @staticmethod
+    def deserialize(obj):
         """Deserialize the spec from a dictionary."""
-        return parse_obj_as(cls, obj)
+        return deserialize_as(Spec, obj)
+
+
+@dataclass(config=StrictConfig)
```
DiamondJoseph, line 119: This spec is still in the wrong place

DiamondJoseph (changes requested):

src/scanspec/core.py
```diff
@@ -104,72 +114,147 @@ def calculate(self) -> int:
...
+            A request for generated scan points
+                spec: Spec
+        This will register `PointsRequest` and `Spec` with the corresponding tagged
+        union or
+            @uses_tagged_union
+            def valid(
+            spec: Spec = Body(..., examples=[_EXAMPLE_SPEC]),
+            ) -> ValidResponse | JSONResponse:
+            The decorator helps pydantic in understanding the core schema for the
+            function
```
DiamondJoseph, line 163: I'd nix this, it's implementation details. Docs are for how/when to use something, not what it does: how it does it should be allowed to change without the docstring needing to be updated.

```suggestion
```

src/scanspec/specs.py
```diff
@@ -506,6 +510,12 @@ def bounded(
         return cls(axis, start, stop, num)
 
 
+"""
+Validation Decorator is requied as we are using custom pydantic core schema validators
+"""
```
DiamondJoseph, line 515:
```suggestion
Defers wrapping function with validate_call until class is fully instantiated
"""
```

DiamondJoseph (approved):

PR #140, opened 2024-08-27
Remove the need for uses_tagged_union deco

coretl:
At present we support people adding to the list of Specs after importing the service. This requires a uses_tagged_union decorator on any BaseModel that references a Spec. We cannot currently think of a use case for Specs being implemented outside of scanspec.specs, so removing this requirement means we can ditch the decorator. It also makes the docs (next PR) easier to write:
- Put your specs in src/scanspec/specs.py
- Release a new version
- Update both blueapi and the scanspec service to use that new version

### DIFF
--- a/src/scanspec/core.py
+++ b/src/scanspec/core.py
@@ -3,14 +3,7 @@ from __future__ import annotations
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from functools import lru_cache
 from inspect import isclass
-from typing import (
-    Any,
-    Generic,
-    Literal,
-    TypeVar,
-    get_origin,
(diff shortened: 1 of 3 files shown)


DiamondJoseph:
Blueapi requires being able to make BaseModels with fields that are Specs/Regions as we create BaseModels for the signature of plan functions. This looks from the tests that remain that this is still supported. 

But if we're acknowledging that all Spec implementations will be in this library we can just make a hardcoded `_Spec_ref = Line | ...`...

DiamondJoseph (approved):

ZohebShaikh (approved):

PR #142, opened 2024-09-13
Update to latest copier and make docs nicer

coretl:


src/scanspec/specs.py
```diff
@@ -1,3 +1,10 @@
+"""`Spec` and its subclasses.
```
callumforrester, line 1: Nit: This is a self-referencing description, same for `regions.py` and `core.py`, it seems a bit redundant, similar to

```python
def get_foo() -> int:
    """
    gets foo
    """
    ...
```

callumforrester (approved):
LGTM

src/scanspec/specs.py
coretl: Ha, better words welcomed!

PR #143, opened 2024-09-13
Put pyright in strict mode

### LINKED ISSUE #144: Enable Strict Mode for Pyright
As discussed with @coretl we would like to use the scanspec repository to prototype a number of modern python correctness tooling including pyright strict mode (see [docs](https://github.com/microsoft/pyright/blob/main/docs/configuration.md)).

## Acceptance Criteria
- Strict mode enabled per https://diamondlightsource.github.io/python-copier-template/main/how-to/strict-mode.html
- `tox -e type-checking` gives no errors

callumforrester:
Fixes #144


coretl:
#142 fixes schema

src/scanspec/core.py
```diff
@@ -1,19 +1,25 @@
...
 import numpy as np
+import numpy.typing as npt
 from pydantic import BaseModel, ConfigDict, Field, GetCoreSchemaHandler
 from pydantic.dataclasses import is_pydantic_dataclass, rebuild_dataclass
 from pydantic_core import CoreSchema
 from pydantic_core.core_schema import tagged_union_schema
 
 __all__ = [
     "if_instance_do",
-    "Axis",
```
coretl, line 16: This was in `__all__` so the docs didn't complain about a missing ref, please can you check the docs still look ok after this?

src/scanspec/regions.py
```diff
@@ -32,6 +33,8 @@
     "find_regions",
 ]
 
+NpMask = npt.NDArray[np.bool]
```
coretl, line 36: BTW I recently discovered you can specify shape too, although most of numpy returns `shape=Any` so it's of limited use at the mo:
```python
ScalarType_co = TypeVar("ScalarType_co", bound=np.generic, covariant=True)
OneDArray = np.ndarray[tuple[int], np.dtype[ScalarType_co]]
TwoDArray = np.ndarray[tuple[int, int], np.dtype[ScalarType_co]]

NPMask = OneDArray[np.bool]
```

src/scanspec/specs.py
```diff
@@ -45,6 +42,12 @@
 #: Can be used as a special key to indicate how long each point should be
 DURATION = "DURATION"
 
+#: A type variable for an `axis_` that can be specified for a scan
+Axis = TypeVar("Axis", covariant=True)
```
coretl, line 46: Why re-implement rather than import?

src/scanspec/specs.py
```diff
@@ -84,34 +89,42 @@ def shape(self) -> tuple[int, ...]:
...
-    def __rmul__(self, other) -> Product[Axis]:
+    def __rmul__(self, other: int) -> Product[Axis]:
         return if_instance_do(other, int, lambda o: Product(Repeat(o), self))
 
-    def __mul__(self, other) -> Product[Axis]:
+    @overload
+    def __mul__(self, other: Spec[Axis]) -> Product[Axis]: ...
+
+    @overload
+    def __mul__(self, other: Spec[OtherAxis]) -> Product[Axis | OtherAxis]: ...
```
coretl, line 99: It's variance like this that makes me think `duration` should be special, otherwise we end up with lots of `Spec[Motor | str]` and `isinstance` checks.

Alternatively we might decide that we always need `Spec[str]` with a lookup, as the case for `Spec[Motor]` in ophyd_async is looking tenuous. One for the scanspec v1 ticket (I realised we never made scanspec 1.0 so might as well use that number).

coretl (approved):

src/scanspec/core.py
callumforrester: Since I agree with https://github.com/bluesky/scanspec/pull/143#discussion_r1759070501, I'm putting this back, however I was planning to manually check the docs and the plotting once you were happy with everything else, so will leave this thread open as a placeholder for that.

src/scanspec/regions.py
callumforrester: Interesting... worth a separate issue to specify shape where possible?

src/scanspec/specs.py
callumforrester: My mistake, I thought it was best practice, but having researched a bit more I agree that importing `TypeVars` is the way to go. Case in point: When I removed these it broke because `Axis` in `core.py` wasn't covariant. I have now fixed and imported both variables.

src/scanspec/specs.py
callumforrester: A few thoughts here:

- `Spec[Motor]` would be very nice for blueapi, if we can get it to work with pydantic, even `Spec[Motor, str]`. 
- If we make time special there might be a way to incorporate it into a `Spec[Motor]` without making it a `Spec[Motor, str]` (or really a `Spec[Motor, Time]`, which would be preferable)
- _However_ we should ask ourselves if the variance is actually a bad thing, the extra verbosity here is just capturing a special case at "compile" time rather than runtime, the special case always existed and was always special, strict mode simply forces us to acknowledge it. In other words it is possible that a bunch of `isinstance` checks really might be the right thing to do here, or it's possible that we can box the specialness into one place by treating time as special somehow. I am not sure, but I don't think we should be prescriptive.

coretl (approved):
Happy for you to merge when you've checked the docs

PR #146, opened 2024-10-09
Ensure minimums are same type as fields

DiamondJoseph:
Json schema generation seems to inconsistently use the minimum arguments with or without casting to the expected type when called from fastapi or not. Ensure that these arguments match the expected type.


ZohebShaikh:
Would it be possible to add a test to verify the typecasting of field values of all the other Pydantic models? 
This would be useful in identifying such cases in the future when similar changes are made.

keithralphs (approved):

DiamondJoseph:
Note that handling appears to be highly inconsistent between anything handling the json blob in a browser (where Javascript appears to truncate 0.0 to 0 regardless of whether the minimum is 0 or 0.0) or any "dumb" client where the form in the definition is respected.

Issue #72, opened 2023-01-04
Use autodoc_pydantic when it supports dataclasses


callumforrester:
Import it as a sphinx plugin after this issue is handled and released:

https://github.com/mansenfranzen/autodoc_pydantic/issues/121

PR #147, opened 2024-10-15
Move autodoc pydantic git dep into ci and devcontainer so we can release

coretl:
Temporary fix to allow release while waiting on a more complete solution for #72


DiamondJoseph (approved):
It's no worse than it already is.

But I don't like the fragility of us depending on a fork that is already behind head and is just a commit hash

callumforrester (approved):

coretl:
I think @paula-mg was thinking about using BaseModel with positional args rather than dataclasses as part of #141 so this may all go away then

coretl:
@DiamondJoseph feel free to make a release when you have your change in

PR #150, opened 2024-10-18
Replace use of deprecated np.bool with dtype alias

DiamondJoseph:


DiamondJoseph:
See: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

In its current condition, loading ScanSpec causes blueapi to fall over when it tries to import the generic `scan` plan

PR #157, opened 2025-01-31
Include generics in discriminated union schemas

### LINKED ISSUE #156: discriminated_union_of_subclasses loses type information when deserialising
## The problem

The `discriminated_union_of_subclasses` decorator is intended to allow
deserialisation a subclass of a base type based on a(n automatically added)
discriminant field at runtime, eg

<details open><summary>Union of fully static subclasses</summary>

```python
@discriminated_union_of_subclasses
class Parent: pass

@dataclass
class ChildOne(Parent):
    foo: int

@dataclass
class ChildTwo(Parent):
    bar: str

assert TypeAdapter(Parent).validate_python({"type": "ChildOne", "foo": "42"}).foo == 42 # note auto conversion to int
assert TypeAdapter(Parent).validate_python({"type": "ChildTwo", "bar": "42"}).bar == "42"
```
</details>

This adds a `type` discriminator field and creates a schema at runtime that
means deserialising `Parent` is roughly equivalent to deserialising `ChildOne |
ChildTwo`.

When everything is fully statically typed, this works as expected and the types
are correct.

When the parent class is generic, this is no longer the case, for instance given
the following system

<details open><summary>Union of generic subclasses</summary>

```python
@discriminated_union_of_subclasses
class GenericParent(Generic[T]): pass

@dataclass
class GenericChild(GenericParent[T]):
    a: T
```

</details>

Deserialising a `GenericParent[int]` (for instance via
`TypeAdapter(GenericParent[int]).validate_python({"type": "GenericChild", "a":
"42"})`)
loses the generic type information and is roughly equivalent to
`TypeAdapter(GenericChild).validate_python({"a": "42"})`. This deserialisation
is successful (in that there are no validation errors) but results in `a` being
a `str` while the type checks treat it as an `int`.

## Difficulty passing type information

The generic type information is available at runtime (it is passed to
`__get_pydantic_core_schema__` but is currently ignored) so while it should be
possible to get better support, fully supporting all use cases is (unless I'm
missing something) surprising complex.

For the sample above where the subclasses have the same type parameters as the
base class and are mapped directly, it is straightforward to pass the type
parameters through.

Even when there are additional parameters such as

```python
@dataclass
class ExtraArgs(GenericParent[T], Generic[T, U]):
    a: T
    b: U
```

the type args can be handled in the same way and passed through to the correct
places in the subclasses in the generated union.

The difficulty comes when the subclasses either partially specify the types of
the `Parent` class or impose additional bounds/constraints.

For instance, given the following class

```python
@dataclass
class IntegerChild(GenericParent[int]):
    a: int
```

the subclass `IntegerChild` is no longer generic but should only be included if
deserialising a `GenericParent[int]`.

Using `TypeVar`s with bounds and/or constraints makes this more complicated
again. A class such as

```python
U = TypeVar('U', bound= ProtocolOne | ProtocolTwo)
@dataclass
class Bounded(GenericParent[U]):
    a: U
```
would require the types used in the generated union used to deserialise the
`GenericParent` to be combined/narrowed to fit the bounds. Including variance
considerations introduces another layer of complexity again.

At this point we wouldn't be far off needing to build a full type checker.

## Possible Solutions

As far as I can see, the current uses of the decorator are only `Region` and
`Spec`. Both are only generic over a single parameter and no subclass introduces
additional parameters or bounds. With this in mind there are potentially
workable approaches.

### Restrict supported subclasses

Restricting subclasses to only those that pass type args straight to the parent
type could work while still supporting the current use cases. This would be
simplest. When subclassing a decorated base class with a more complicated class,
it could warn and then ignore that type when deserialising.

If the 'extra args' version above was required for some reason, these could also
be supported at the cost of slight additional complexity in forwarding the args
to the correct places if they are ordered differently, eg in `class
Disordered(Parent[T, U], Generic[W, U, T]): pass`.

### Replace the decorator with an alias

If we know up front what subtypes are going to be available, we can make the
union of subclasses up front and remove the need for the decorator.

## Acceptance Criteria
- Types are deserialised to the specified type or an explicit error/warning is raised
  that the type is not supported along with documentation for what is required for
  the decorator to work.

tpoliaw:
- **Restrict subclasses of typed unions**
- **Allow arbitrary types to be included in 'StrictConfig' types**

Fixes #156

### Instructions to reviewer on how to test:
1. Create set of subclasses of a decorated base class:
  ```python
  @discriminated_union_of_subclasses
  class Base(Generic[T]):
      pass

  @dataclass
  class Child(Base[T]):
      foo: T
  ```
2. Try and deserialise a serialised `Child` as a `Base`:
  ```python
  one = TypeAdapter(Base[int]).validate_python({"type": "Child", "foo": "42"})
  assert one.foo == 42 # <== 42 instead of "42" as type should be converted
  ```

### Checks for reviewer
- [ ] Would the PR title make sense to a user on a set of release notes


src/scanspec/core.py
```diff
@@ -20,6 +25,18 @@
...
+except ImportError:
+    # function added to stdlib in 3.12
+    def get_original_bases(cls: type, /) -> tuple[Any, ...]:
+        try:
+            return cls.__dict__.get("__orig_bases__", cls.__bases__)
+        except AttributeError:
+            raise TypeError(
+                f"Expected an instance of type, not {type(cls).__name__!r}"
+            ) from None
+
```
DiamondJoseph, line 39: Can we use an explicit version check? It's easier to see when to remove when ScanSpec no longer supports <3.12

```python
if sys.version_info < (3, 12):
     ...
else:
     ...
```

src/scanspec/core.py
```diff
@@ -174,13 +200,41 @@ def _rebuild(cls_or_func: Callable[..., T]) -> None:
...
+def _parameters(possibly_generic: type) -> tuple[Any,...]:
+    return getattr(possibly_generic, '__parameters__', ())
+
+def _compatible_types(left: TypeVar, right: TypeVar) -> bool:
+    return (
+            left.__bound__ == right.__bound__
+            and left.__constraints__ == right.__constraints__
+            and left.__covariant__ == right.__covariant__
+            and left.__contravariant__ == right.__contravariant__
+            )
```
DiamondJoseph, line 237: Could left.__ be a subset of right.__?

```
class Super:
class Sub(Super):

@discriminated...
class Parent(Generic[Super]):
class Child(Parent[Sub], Generic[Sub])
```

e.g. for class ReadableMovable(Movable)?

DiamondJoseph (approved):
A nit and a possible edge case

src/scanspec/core.py
tpoliaw: You can't pass concrete types to `Generic` as in this example but yes, this is definitely restricting the supported cases beyond what is possible in theory. It turns out to be difficult to determine what should be passed without re-implementing an entire type-checker. There are a couple of examples in #156 and a few in a previous version of this commit where I tried to support a wider range of options but until we have a use case for them I'd prefer a well defined restriction instead of an array of unsupported edge-cases.

src/scanspec/core.py
DiamondJoseph: I think (once you find one) having a search term e.g. `sys.version_info` to find the rest can save time.

tpoliaw: Done, slight downside now is that the function isn't type-checked if it's run on a 3.12+ system

callumforrester:
Delete uncovered lines as discussed

callumforrester:
Should consider adding a codecov config like [ophyd-async's](https://github.com/bluesky/ophyd-async/blob/main/.codecov.yml) for consistency

DiamondJoseph:
Rebased to pick up codecov changes

PR #169, opened 2025-06-20
167 separate duration out

LuisFSegalla:
This PR has two main objectives:

1. Renaming classes to avoid confusing names.
2. Remove **DURATION** from the list of axis.

The PR is still incomplete but it will allow us to discuss important points for the development. 

### Issues to be discussed:

1. Making the Dimensions class accept `None` for the midpoint argument, we will have to handle the check of midpoints in parts of the code where previously it was expected, at least the **midpoints** attribute to exist.
2. Should the **duration** spec have only the duration of each frame and get the amount of frames from the other axis or should it also include a **num** argument

### DIFF
--- a/src/scanspec/core.py
+++ b/src/scanspec/core.py
@@ -289,6 +289,30 @@ OtherAxis = TypeVar("OtherAxis")
 AxesPoints = dict[Axis, npt.NDArray[np.float64]]
 
 
+# @dataclass
+class ScanSlice(Generic[Axis]):
+    """Generalization of the Dimensions class.
+
+    Only holds information but no methods to handle it.
+    """
+
+    def __init__(
+        self,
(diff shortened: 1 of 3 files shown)


coretl:
> * Making the Dimensions class accept `None` for the midpoint argument, we will have to handle the check of midpoints in parts of the code where previously it was expected, at least the **midpoints** attribute to exist.

Thinking about it, what happens if `midpoints` must exist, but is allowed to be the empty dictionary. Does that work?

> * Should the **duration** spec have only the duration of each frame and get the amount of frames from the other axis or should it also include a **num** argument

Duration only, we can do `Duration(duration) * Static(5)` if we want multiples

coretl:
Superceded by #172

PR #175, opened 2025-06-27
Add Slice class and update consume method

LuisFSegalla:
This PR serves as a middle step to update way we consume a **Path**.
The new **Slice** class contain only information about the scan and it decouples the **DURATION** axis, if present, from the **midpoints** information. This should make checks about the Path slightly easier.

An auxiliary function (**stack2dimension**) was added to guarantee that other parts of the code that used **Path.consume()** still worked without having to change a lot of the code.

### DIFF
--- a/docs/how-to/iterate-a-spec.rst
+++ b/docs/how-to/iterate-a-spec.rst
@@ -59,12 +59,12 @@ as you need. You can also give a maximum size to `Path.consume()`
 >>> len(path)
 3
 >>> chunk = path.consume(2)
->>> len(chunk)
+>>> len(chunk.midpoints["x"])
 2
 >>> len(path)
 1
 >>> chunk = path.consume(2)
->>> len(chunk)
+>>> len(chunk.midpoints["x"])
 1
 >>> len(path)
 0
@@ -75,7 +75,7 @@ You can also use this method to only run a subset of the scan:
 >>> len(path)
 1
 >>> chunk = path.consume()
->>> len(chunk)
+>>> len(chunk.midpoints["x"])
 1
 >>> len(path)
 0
--- a/src/scanspec/core.py
+++ b/src/scanspec/core.py
@@ -50,7 +50,11 @@ __all__ = [
     "Midpoints",
     "discriminated_union_of_subclasses",
     "StrictConfig",
+    "DURATION",
 ]
+
+#: Can be used as a special key to indicate how long each point should be
+DURATION = "DURATION"
 
 #: Used to ensure pydantic dataclasses error if given extra arguments
 StrictConfig: ConfigDict = {"extra": "forbid", "arbitrary_types_allowed": True}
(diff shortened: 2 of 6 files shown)


docs/how-to/iterate-a-spec.rst
```diff
@@ -59,12 +59,12 @@ as you need. You can also give a maximum size to `Path.consume()`
 >>> len(path)
 3
 >>> chunk = path.consume(2)
->>> len(chunk)
+>>> len(chunk.midpoints["x"])
```
coretl, line 62: Actually, this is probably a valid thing to support, can we put the `__len__` method onto the Slice please?

src/scanspec/core.py
```diff
@@ -582,7 +671,7 @@ def squash_frames(
     """
     path = Path(stack)
```
coretl, line 672: I guess we can delete this line now?
```suggestion
```

src/scanspec/core.py
```diff
@@ -288,6 +292,29 @@ def if_instance_do(x: C, cls: type[C], func: Callable[[C], T]) -> T:
...
+        lower: AxesPoints[Axis] | None = None,
+        upper: AxesPoints[Axis] | None = None,
+        duration: DurationArray | None = None,
+    ):
+        self.axes = axes
+        self.midpoints = midpoints
+        self.lower = lower
+        self.upper = upper
+        self.gap = gap
+        self.duration = duration
```
coretl, line 315: Any reason not to make this a dataclass?

src/scanspec/core.py
```diff
@@ -288,6 +292,29 @@ def if_instance_do(x: C, cls: type[C], func: Callable[[C], T]) -> T:
...
+    """
+
+    def __init__(
+        self,
+        axes: list[Axis],
+        midpoints: AxesPoints[Axis],
+        gap: GapArray,
+        lower: AxesPoints[Axis] | None = None,
+        upper: AxesPoints[Axis] | None = None,
+        duration: DurationArray | None = None,
```
coretl, line 308: I don't think any of these can be None, as they are always expanded by `Dimension` and passed to this function

```suggestion
        lower: AxesPoints[Axis],
        upper: AxesPoints[Axis],
        duration: DurationArray,
```

docs/how-to/iterate-a-spec.rst
LuisFSegalla: Done!

src/scanspec/core.py
LuisFSegalla: That makes sense! But I'm wondering, we're still able to class `Path.consume()` into a Dimension that doesn't have the `duration` defined, so I think we could keep the option for `duration` to be None by default.

src/scanspec/core.py
```diff
@@ -482,6 +507,68 @@ def _merge_frames(
...
 
+def stack2dimension(
+    dimensions: list[Dimension[Axis]],
+    indices: npt.NDArray[np.signedinteger] | None = None,
+    lengths: npt.NDArray[np.signedinteger] | None = None,
+) -> Dimension[Axis]:
+    if lengths is not None:
+        lengths = lengths
+    else:
+        lengths = np.array([len(f) for f in dimensions])
```
coretl, line 518: Nit: in general I prefer
```suggestion
    if lengths is None:
        lengths = np.array([len(f) for f in dimensions])
```

tests/test_service.py
```diff
@@ -49,7 +49,10 @@ def test_subsampling(client: TestClient) -> None:
         "total_frames": 25,
         "returned_frames": 8,
         "format": "FLOAT_LIST",
-        "midpoints": {"x": [0.0, 0.0, 10.0, 10.0], "y": [0.0, 10.0, 0.0, 10.0]},
+        "midpoints": {
+            "x": [0.0, 0.0, 0.0, 0.0, 0.0, 2.5, 2.5, 2.5],
+            "y": [0.0, 2.5, 5.0, 7.5, 10, 0.0, 2.5, 5.0],
+        },
     }
```
coretl, line 56: Why is this different?

src/scanspec/core.py
LuisFSegalla: Done!

coretl (approved):

PR #176, opened 2025-07-10
172 create duration spec and update fly and step

LuisFSegalla:
This Pull Request aims at adding a new spec class `Duration` that contains only a `Float`. This class is now used in `fly` and `step` as  a new `Dimension` that is `zipped` with the other ones.

### DIFF
--- a/tests/test_specs.py
+++ b/tests/test_specs.py
@@ -67,7 +67,7 @@ def test_two_point_stepped_line() -> None:
 
 
 def test_two_point_fly_line() -> None:
-    inst = Fly(ConstantDuration(constant_duration=0.1, spec=Line(x, 0, 1, 2)))
+    inst = Fly(0.1 @ Line(x, 0, 1, 2))
     (dim,) = inst.calculate()
     assert dim.midpoints == {
         x: approx([0, 1]),
@@ -98,14 +98,14 @@ def test_empty_dimension() -> None:
 
 
 def test_concat() -> None:
-    dim1 = Fly(spec=ConstantDuration(constant_duration=1, spec=Line("x", 0, 1, 2)))
+    dim1 = Fly(1.0 @ Line("x", 0, 1, 2))
     dim2 = Line("x", 3, 4, 2)
 
     with pytest.raises(ValueError) as msg:
         Concat(dim1, dim2)
     assert "Only one of left and right defines a duration" in str(msg.value)
 
-    dim2 = Fly(spec=ConstantDuration(constant_duration=1, spec=Line("x", 3, 4, 2)))
+    dim2 = Fly(1.0 @ Line("x", 3, 4, 2))
 
     spec = Concat(dim1, dim2)
 
@@ -113,7 +113,7 @@ def test_concat() -> None:
 
     # Check that concat on the Dimension class works as expected
     (dim1,) = Line("x", 3, 4, 2).calculate()
-    (dim2,) = ConstantDuration(1, Line("x", 3, 4, 2)).calculate()
+    (dim2,) = (1.0 @ Line("x", 3, 4, 2)).calculate()
 
     with pytest.raises(ValueError) as msg:
         dim1.concat(dim2)
@@ -123,8 +123,8 @@ def test_concat() -> None:
 
 
 def test_zip() -> None:
-    dim1 = Fly(spec=ConstantDuration(constant_duration=1, spec=Line("x", 0, 1, 2)))
-    dim2 = Fly(spec=ConstantDuration(constant_duration=2, spec=Line("y", 3, 4, 2)))
+    dim1 = Fly(1.0 @ Line("x", 0, 1, 2))
+    dim2 = Fly(2.0 @ Line("y", 3, 4, 2))
 
     with pytest.raises(ValueError) as cm:
         Zip(dim1, dim2)
@@ -242,7 +242,7 @@ def test_squashed_multiplied_snake_scan() -> None:
     inst = Line(z, 1, 2, 2) * Squash(
         Line(y, 1, 2, 2)
         * ~Line.bounded(x, 3, 7, 2)
-        * ConstantDuration(9, Repeat[str](2, gap=False))  # until #177
+        * (9.0 @ Repeat[str](2, gap=False))  # until #177
     )
     assert inst.axes() == [z, y, x]
     (dimz, dimxyt) = inst.calculate()
@@ -279,9 +279,7 @@ def test_product_snaking_lines() -> None:
 
 def test_product_duration() -> None:
     with pytest.raises(ValueError) as msg:
-        _ = Fly(ConstantDuration(1, Line(y, 1, 2, 3))) * Fly(
-            ConstantDuration(1, ~Line(x, 0, 1, 2))
-        )
+        _ = Fly(1.0 @ Line(y, 1, 2, 3)) * Fly(1.0 @ ~Line(x, 0, 1, 2))
     assert "Outer axes defined a duration" in str(msg.value)
 
 
@@ -296,13 +294,11 @@ def test_concat_lines() -> None:
 
     # Test concating one Spec with duration and another one without
     with pytest.raises(ValueError) as msg:
-        Concat(ConstantDuration(1, Line(x, 0, 1, 2)), Line(x, 1, 2, 3))
+        Concat((1.0 @ Line(x, 0, 1, 2)), Line(x, 1, 2, 3))
     assert "Only one of left and right defines a duration" in str(msg.value)
 
     # Variable duration concat
-    spec = Concat(
-        ConstantDuration(1, Line(x, 0, 1, 2)), ConstantDuration(2, Line(x, 1, 2, 3))
-    )
+    spec = Concat(1.0 @ Line(x, 0, 1, 2), 2.0 @ Line(x, 1, 2, 3))
     assert spec.duration() == VARIABLE_DURATION
 
 
@@ -383,12 +379,7 @@ def test_rect_region_difference() -> None:
         Line(y, 1, 3, 5) * Line(x, 0, 2, 3) & Rectangle(x, y, 0, 1, 1.5, 2.2)
     ) - Rectangle(x, y, 0.5, 1.5, 2, 2.5)
 
-    inst = Fly(
-        ConstantDuration(
-            0.1,
-            spec,
-        )
-    )
+    inst = Fly(0.1 @ spec)
     assert inst.axes() == [y, x]
     (dim,) = inst.calculate()
     assert dim.midpoints == {
@@ -623,11 +614,11 @@ def test_shape(spec: Spec[Any], expected_shape: tuple[int, ...]):
 
 
 def test_constant_duration():
-    spec1 = Fly(ConstantDuration(spec=Line("x", 0, 1, 2), constant_duration=1))
-    spec2 = ConstantDuration(2, Line("x", 0, 1, 2))
+    spec1 = Fly(1.0 @ Line("x", 0, 1, 2))
+    spec2 = 2.0 @ Line("x", 0, 1, 2)
 
     with pytest.raises(ValueError) as msg:
-        ConstantDuration(2, spec1)
+        2.0 @ spec1  # type: ignore
     assert f"{spec1} already defines a duration" in str(msg.value)
 
     with pytest.raises(ValueError) as msg:
@@ -676,11 +667,9 @@ def test_step():
 
 @pytest.mark.filterwarnings("ignore:get_constant_duration")
 def test_get_constant_duration():
-    spec = Fly(ConstantDuration(1, Line("x", 0, 1, 4))).calculate()
+    spec = Fly(1.0 @ Line("x", 0, 1, 4)).calculate()
     assert get_constant_duration(spec) == 1
 
-    spec = Concat(
-        ConstantDuration(1, Line(x, 0, 1, 2)), ConstantDuration(2, Line(x, 1, 2, 3))
-    ).calculate()
+    spec = Concat(1.0 @ Line(x, 0, 1, 2), 2.0 @ Line(x, 1, 2, 3)).calculate()
 
     assert get_constant_duration(spec) is None
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -123,7 +123,7 @@ def test_plot_1D_line_snake_repeat() -> None:
 
 def test_plot_1D_step() -> None:
     runner = CliRunner()
-    spec = 'ConstantDuration(0.1,Line("x", 1, 2, 2))'
+    spec = '0.1 @ Line("x", 1, 2, 2)'
     with patch("scanspec.plot.plt.show"):
         result = runner.invoke(cli.cli, ["plot", spec])
     assert result.stdout == ""


src/scanspec/core.py
```diff
@@ -375,13 +377,16 @@ def __init__(
...
+        if len(midpoints) > 0:
+            # Check all lengths are the same
+            lengths = {
+                len(arr)
+                for d in (self.midpoints, self.lower, self.upper)
+                for arr in d.values()
+            }
+            lengths.add(len(self.gap))
+        else:
+            lengths = {len(self.duration)} if self.duration is not None else {0}
```
coretl, line 389: If duration and midpoints both exist we need to check that

```suggestion
        # Check all lengths are the same
        lengths = {
            len(arr)
            for d in (self.midpoints, self.lower, self.upper)
            for arr in d.values()
        }
        lengths.add(len(self.gap))
        if self.duration is not None:
            lengths.add(len(self.duration))
```

src/scanspec/core.py
```diff
@@ -483,13 +501,27 @@ def zip_gap(gaps: Sequence[GapArray]) -> GapArray:
             # gap[i] = self.gap[i] | other.gap[i]
             return np.logical_or.reduce(gaps)
 
-        return _merge_frames(self, other, dict_merge=zip_dict, gap_merge=zip_gap)
+        def zip_duration(
+            durations: Sequence[DurationArray | None],
+        ) -> DurationArray | None:
+            # assert len(durations) == 1, "Can't have more than one durations array"
+            return durations[-1]
```
coretl, line 508: Need to check that only one duration array is given:

```suggestion
            # We catch the case of all None in merge_frames, so only handle multiple durations here
            specified_durations = [d for d in durations if d is not None]
            if len(specified_durations) != 1:
                raise ValueError("Can't have more than one durations array")
            return specified_durations[0]
```

Can we have a test for this too please

src/scanspec/service.py
```diff
@@ -223,7 +223,7 @@ def gap(
     dims = spec.calculate()  # Grab dimensions from spec
     path = Path(dims)  # Convert to a path
     gap = list(path.consume().gap)
-    return GapResponse(gap)
+    return GapResponse(gap)  # type: ignore
```
coretl, line 226: Why is type ignore needed?

src/scanspec/specs.py
```diff
@@ -244,6 +244,14 @@ def calculate(  # noqa: D102
             # Take the 0th element N times to make a repeated Dimension object
             indices = np.zeros(len(frames_left[-1]), dtype=np.int8)
             repeated = frames_right[0].extract(indices)
+            # Check if frames_right is a Duration type frame
+            if (
+                len(repeated.midpoints) == 0
+                and repeated.duration is not None
+                and len(repeated.duration) != 0
+            ):
+                repeated.gap = np.full(len(repeated.duration), False)
```
coretl, line 253: Why is this needed? If `Dimension.extract` doesn't do the right thing then shouldn't it be added there?

src/scanspec/specs.py
```diff
@@ -549,6 +557,38 @@ def bounded(
...
+    def axes(self) -> list[Axis]:  # noqa: D102
+        return []
+
+    @classmethod
+    def redefine(
+        cls: type[Duration[Any]],
+        duration: float = Field(description="The duration of each static point"),
+        num: int = Field(ge=1, description="Number of frames to produce", default=1),
+    ) -> Duration[str]:
+        return Duration(duration, num)
```
coretl, line 576: Don't need this function
```suggestion
```

src/scanspec/specs.py
```diff
@@ -549,6 +557,38 @@ def bounded(
 Line.bounded = validate_call(Line.bounded)  # type:ignore
 
 
+@dataclass(config=StrictConfig)
+class Duration(Spec[Axis]):
+    """A special spec used to hold information about the duration of each frame."""
+
+    value: float = Field(description="The value at each point")
```
coretl, line 564: Slight preference for `duration` rather than `value`?

src/scanspec/specs.py
```diff
@@ -714,7 +754,7 @@ def fly(spec: Spec[Axis], duration: float) -> Spec[Axis | str]:
         spec = fly(Line("x", 1, 2, 3), 0.1)
 
     """
-    return spec.zip(Static.duration(duration))
+    return spec.zip(Duration(duration, 1))
```
coretl, line 757: Nit:

```suggestion
    return spec.zip(Duration(duration))
```

src/scanspec/specs.py
```diff
@@ -733,7 +773,7 @@ def step(spec: Spec[Axis], duration: float, num: int = 1) -> Spec[Axis | str]:
         spec = step(Line("x", 1, 2, 3), 0.1)
 
     """
-    return spec * Static.duration(duration, num)
+    return spec * Duration(duration, 1)
```
coretl, line 776: ditto

tests/test_specs.py
```diff
@@ -63,7 +63,10 @@ def test_two_point_stepped_line() -> None:
     inst = step(Line(x, 0, 1, 2), 0.1)
     dimx, dimt = inst.calculate()
     assert dimx.midpoints == dimx.lower == dimx.upper == {x: approx([0, 1])}
-    assert dimt.midpoints == dimt.lower == dimt.upper == {DURATION: approx([0.1])}
+    assert dimt.midpoints == dimt.lower == dimt.upper == {}
+    assert dimt.duration == approx(
+        [0.1]
+    )  # Don't think this is how it's supposed to work
```
coretl, line 69: I don't understand the comment, this looks right to me?

coretl (changes requested):

src/scanspec/specs.py
LuisFSegalla: Removed

src/scanspec/specs.py
LuisFSegalla: As discussed in person, I updated the `__init__` method of the `Dimension` class to update `.gap` to a consistent value making this part of the code unnecessary.

src/scanspec/core.py
LuisFSegalla: Done!

LuisFSegalla:
Changed this PR to be a draft for the time being as I'm not sure this is the right way to implement what was discussed with @coretl.

src/scanspec/core.py
```diff
@@ -458,7 +478,21 @@ def concat_gap(gaps: Sequence[GapArray]) -> GapArray:
...
             return g
 
-        return _merge_frames(self, other, dict_merge=concat_dict, gap_merge=concat_gap)
+        def concat_duration(
+            durations: Sequence[DurationArray | None],
+        ) -> DurationArray | None:
+            # Check if there are more than one duration being zipped
+            specified_durations = [d for d in durations if d is not None]
+            d = np.concatenate(specified_durations)
+            return d
```
coretl, line 487: I think we should require they are all not None, it doesn't make sense to concatentate them otherwise

```suggestion
            # We will be passed a sequence of durations where at least one
            # is not None. We require that they are all not None
            if any(d is None for d in durations):
                raise ValueError("Can't concatenate dimensions unless all or none provide durations")
            return np.concatenate(durations)
```

src/scanspec/core.py
```diff
@@ -423,7 +430,20 @@ def extract_gap(gaps: Iterable[GapArray]) -> GapArray | None:
...
             return None
 
-        return _merge_frames(self, dict_merge=extract_dict, gap_merge=extract_gap)
+        def extract_duration(
+            durations: Iterable[DurationArray | None],
+        ) -> DurationArray | None:
+            for d in durations:
+                if d is not None:
+                    return d[dim_indices]
+                return None
```
coretl, line 439: For consistency with the above functions

```suggestion
            return None
```

src/scanspec/specs.py
```diff
@@ -149,6 +162,19 @@ class Product(Spec[Axis]):
...
+            return None
+        else:
+            if self.outer.duration() == self.inner.duration:
+                return self.inner.duration()
+            elif self.outer.duration() is None:
+                return self.inner.duration()
+            elif self.inner.duration():
+                return self.outer.duration()
+            else:
+                return VARIABLE_DURATION
```
coretl, line 176: I think we should say that either the outer or the inner produces duration, not both.

```suggestion
        outer, inner = self.outer.duration(), self.inner.duration()
        if outer is not None and inner is not None:
            raise ValueError("Both outer and inner define a duration")
        return inner if inner is not None else outer
```

src/scanspec/specs.py
```diff
@@ -229,6 +258,19 @@ class Zip(Spec[Axis]):
...
+            return None
+        else:
+            if self.left.duration() == self.right.duration:
+                return self.right.duration()
+            elif self.left.duration() is None:
+                return self.right.duration()
+            elif self.right.duration():
+                return self.left.duration()
+            else:
+                return VARIABLE_DURATION
```
coretl, line 272:
```suggestion
        left, right = self.left.duration(), self.right.duration()
        if left is not None and right is not None:
            raise ValueError("Both left and right define a duration")
        return left if left is not None else right
```

src/scanspec/specs.py
```diff
@@ -299,6 +342,12 @@ class Mask(Spec[Axis]):
     def axes(self) -> list[Axis]:  # noqa: D102
         return self.spec.axes()
 
+    def duration(self):  # noqa: D102
+        if self.spec.duration() is None:
+            return None
+        else:
+            return self.spec.duration()
```
coretl, line 349:
```suggestion
        return self.spec.duration()
```

src/scanspec/specs.py
```diff
@@ -404,6 +456,19 @@ def axes(self) -> list[Axis]:  # noqa: D102
...
+            return None
+        else:
+            if self.left.duration() == self.right.duration:
+                return self.right.duration()
+            elif self.left.duration() is None:
+                return self.right.duration()
+            elif self.right.duration():
+                return self.left.duration()
+            else:
+                return VARIABLE_DURATION
```
coretl, line 470: Both or neither should define a duration

```suggestion
        left, right = self.left.duration(), self.right.duration()
        if left == right:
            # They are producing the same duration, whatever it is
            return left
        elif left is None or right is None:
            # They aren't both None, but if one is then raise
            raise ValueError("Only one of left and right defines a duration")
        else:
            # They both exist, but are different, so are variable
            return VARIABLE_DURATION
```

coretl (changes requested):

src/scanspec/specs.py
```diff
@@ -64,13 +62,28 @@ class Spec(Generic[Axis]):
...
 
+    def duration(self) -> float | None | Literal["VARIABLE_DURATION"]:
+        """Returns the duration of each scan point.
+
+        Return value will be one of:
+        - ``None``: No duration defined
+        - ``float``: A constant duration for each point
+        - `VARIABLE_DURATION`: A different duration for each point
+        """
+        raise NotImplementedError(self)
```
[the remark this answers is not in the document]
LuisFSegalla: I tried making `duration` into a `@property` but stumbled into a few problems. It still complained if I tried changing `constant_duration` to just `duration` as saying that it is `obscured by a declaration of the same name`. 
It also made it so that the initially check we do in the `__post_init__` method doesn't work with the duration as it's not a method.

src/scanspec/core.py
```diff
@@ -370,6 +371,10 @@ def __init__(
                 )
             ]
             self.gap = np.logical_or.reduce(axes_gap)
+        # If only duratiotn is provided we need to make gap have the same shape
+        # as the duration provided for the __len__ method
+        elif gap is None and self.duration is not None and len(self.midpoints) == 0:
+            self.gap = np.full(len(self.duration), False)
```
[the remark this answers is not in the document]
LuisFSegalla: Done! It now raises a `ValueError` if it reaches the `else` statement.

src/scanspec/core.py
```diff
@@ -506,6 +557,9 @@ def _merge_frames(
         upper=dict_merge([fs.upper for fs in stack])
         if any(fs.midpoints is not fs.upper for fs in stack)
         else None,
+        duration=duration_merge([fs.duration for fs in stack])  # type: ignore | not sure
```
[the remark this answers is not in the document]
LuisFSegalla: Done!

src/scanspec/specs.py
LuisFSegalla: Done!

src/scanspec/specs.py
LuisFSegalla: Done!

src/scanspec/specs.py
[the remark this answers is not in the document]
coretl: ok, let's leave it as it is then

tests/test_specs.py
```diff
@@ -61,28 +51,26 @@ def test_two_point_line() -> None:
...
 def test_two_point_stepped_line() -> None:
     inst = step(Line(x, 0, 1, 2), 0.1)
-    dimx, dimt = inst.calculate()
-    assert dimx.midpoints == dimx.lower == dimx.upper == {x: approx([0, 1])}
-    assert dimt.midpoints == dimt.lower == dimt.upper == {DURATION: approx([0.1])}
-    assert inst.frames().gap == ints("11")
+    (dim, dimt) = inst.calculate()
+    assert dim.midpoints == {x: approx([0, 1])}
+    assert dim.gap == ints("10")
+    assert dimt.duration == approx([0.1, 0.1])
```
coretl, line 57: This should be a single point

```suggestion
    assert dimt.duration == approx([0.1])
```

tests/test_specs.py
```diff
@@ -194,19 +182,28 @@ def test_squashed_product() -> None:
 
 
 def test_squashed_multiplied_snake_scan() -> None:
-    inst = Line(z, 1, 2, 2) * Squash(
-        Line(y, 1, 2, 2) * ~Line.bounded(x, 3, 7, 2) * Static.duration(9, 2)
+    inst = ConstantDuration(
+        Line(z, 1, 2, 2) * Squash(Line(y, 1, 2, 2) * ~Line.bounded(x, 3, 7, 2)),
+        9,
+        fly=True,
```
coretl, line 188: The original test should now be translated to:

```suggestion
    inst = Line(z, 1, 2, 2) * Squash(
        Line(y, 1, 2, 2) * ~Line.bounded(x, 3, 7, 2) * ConstantDuration(Repeat(2), 9, fly=False)
```

tests/test_specs.py
```diff
@@ -548,61 +549,3 @@ def test_multiple_statics_with_grid():
...
-    with pytest.raises(
-        AssertionError, match=re.escape("Zipping would overwrite axes ['DURATION']")
-    ):
-        spec.calculate()
-    spec = (  # TODO: refactor when https://github.com/bluesky/scanspec/issues/90
-        Static.duration(0.1) * Line.bounded(DURATION, 0, 0, 2)
-    )
-    frames = spec.calculate()
-    assert len(frames) == 2
-    assert get_constant_duration(frames) is None
```
coretl, line 608: Please could you add a test that triggers each of the `raise` statements that you have added

coretl (commented):
Hmm, I'm still a little unsure about whether we've got the typing right, but if you do these bits then we can explore in a different ticket

tests/test_specs.py
LuisFSegalla: Done!

LuisFSegalla:
It was decided that in a `Produect` only the `inner` spec can define the duration.
We'll try to default `spec` in `ConstantDuration` to None and move it to the second argument and see if typing will complain.
Change `ConsntantDuration.calculate` to create a new empty `Dimension` with only the duration if no `spec` was provided.
Change `ConsntatDuration.calculate` if `fly == False` to call `self.spec.calculate(bounds=False)` and set duration on the last Dimension, changing the lenght to be based on `.gap` not `.midpoints`

coretl (commented):
Found a few more...

src/scanspec/specs.py
```diff
@@ -549,6 +603,56 @@ def bounded(
...
+            raise ValueError(f"{self.spec} already defines a duration")
+        return self.constant_duration
+
+    def calculate(  # noqa: D102
+        self, bounds: bool = True, nested: bool = False
+    ) -> list[Dimension[Axis]]:
+        if self.spec:
+            dimensions = (
+                self.spec.calculate() if self.fly else self.spec.calculate(bounds=False)
+            )
```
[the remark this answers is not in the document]
coretl: It means that `ConstantDuration.calculate(bounds)` calls `self.spec.calculate(bounds)` and the `bounds=False` unless you have a `Fly` spec somewhere in the chain which passes through `bounds=True`.

tests/test_specs.py
```diff
@@ -194,19 +182,21 @@ def test_squashed_product() -> None:
...
     for d in dimxyt.midpoints, dimxyt.lower, dimxyt.upper:
         assert d == {
             x: approx([4, 4, 6, 6, 6, 6, 4, 4]),
             y: approx([1, 1, 1, 1, 2, 2, 2, 2]),
-            DURATION: approx([9, 9, 9, 9, 9, 9, 9, 9]),
         }
+    assert dimxyt.duration == approx([9, 9, 9, 9, 9, 9, 9, 9])
     assert dimz.midpoints == dimz.lower == dimz.upper == {z: approx([1, 2])}
-    assert inst.frames().gap == ints("1010101010101010")
+    assert inst.frames().gap == ints("1111111111111111")
```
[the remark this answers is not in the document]
LuisFSegalla: Done!

tests/test_specs.py
```diff
@@ -61,28 +51,26 @@ def test_two_point_line() -> None:
 
 def test_two_point_stepped_line() -> None:
     inst = step(Line(x, 0, 1, 2), 0.1)
-    dimx, dimt = inst.calculate()
-    assert dimx.midpoints == dimx.lower == dimx.upper == {x: approx([0, 1])}
-    assert dimt.midpoints == dimt.lower == dimt.upper == {DURATION: approx([0.1])}
-    assert inst.frames().gap == ints("11")
+    (dim,) = inst.calculate()
+    assert dim.midpoints == {x: approx([0, 1])}
```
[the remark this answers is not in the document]
LuisFSegalla: Done!

tests/test_specs.py
```diff
@@ -30,16 +29,6 @@ def ints(s: str) -> Any:
     return approx([int(t) for t in s])
 
 
-def test_one_point_duration() -> None:
-    duration = Static.duration(1.0)
```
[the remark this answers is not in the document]
LuisFSegalla: Done!

src/scanspec/specs.py
```diff
@@ -549,6 +603,56 @@ def bounded(
...
+            for d in dimensions:
+                if d.duration is not None and d.duration != self.duration:
+                    raise ValueError(self)
```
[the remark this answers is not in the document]
LuisFSegalla: After pushing the code I noticed that we'll never reach this condition, because if we try to create a `ConstantDuration` spec passing a spec that contains a `duration` we would raise an error on the `__post_init__` check.

pyproject.toml
```diff
@@ -12,7 +12,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 description = "Specify step and flyscan paths in a serializable, efficient and Pythonic way"
-dependencies = ["numpy", "click>=8.1", "pydantic>=2.0"]
+dependencies = ["numpy", "click>=8.1", "pydantic>=2.0", "Pillow==11.0.0"]
```
LuisFSegalla, line 15: Move this to the dev dependencies

src/scanspec/core.py
```diff
@@ -349,18 +346,22 @@ def __init__(
...
         self.upper = upper or midpoints
+        self.duration = duration
         if gap is not None:
             #: Whether there is a gap between this frame and the previous. First
             #: element is whether there is a gap between the last frame and the first
             self.gap = gap
-        else:
+        # If midpoints are provided and we don't have a gap array
+        # calculate it based on the midpoints
+        elif gap is None and len(self.midpoints) > 0:
```
LuisFSegalla, line 364:
```suggestion
        elif self.midpoints:
```

src/scanspec/specs.py
```diff
@@ -64,15 +63,30 @@ class Spec(Generic[Axis]):
...
 
+    def duration(self) -> float | None | Literal["VARIABLE_DURATION"]:
+        """Returns the duration of each scan point.
+
+        Return value will be one of:
+        - ``None``: No duration defined
+        - ``float``: A constant duration for each point
+        - `VARIABLE_DURATION`: A different duration for each point
+        """
+        raise NotImplementedError(self)
```
LuisFSegalla, line 86: Make default value return None and remove implementations that do it

src/scanspec/specs.py
```diff
@@ -253,7 +283,7 @@ def calculate(  # noqa: D102
         padded_right: list[Dimension[Axis] | None] = [None] * npad
         # Mypy doesn't like this because lists are invariant:
```
LuisFSegalla, line 284: Remove comment

src/scanspec/specs.py
```diff
@@ -549,6 +603,66 @@ def bounded(
...
+    spec: Spec[Axis] = Field(description="Spec contaning the path to be followed")
+
+    def axes(self) -> list[Axis]:  # noqa: D102
+        return self.spec.axes()
+
+    def duration(self):  # noqa: D102
+        return self.spec.duration()
+
+    def calculate(  # noqa: D102
+        self, bounds: bool = True, nested: bool = False
```
LuisFSegalla, line 619: Default still bounds=False

src/scanspec/specs.py
```diff
@@ -700,24 +801,7 @@ def spaced(
 Spiral.spaced = validate_call(Spiral.spaced)  # type:ignore
 
 
-def fly(spec: Spec[Axis], duration: float) -> Spec[Axis | str]:
```
LuisFSegalla, line 703: Don't remove it, but deprecate it

tests/test_specs.py
```diff
@@ -94,6 +82,50 @@ def test_many_point_line() -> None:
...
+
+    (dim2,) = Fly(
+        spec=ConstantDuration(constant_duration=2, spec=Line("y", 3, 4, 2))
+    ).calculate()
+
+    with pytest.raises(ValueError):
+        dim1.zip(dim2)
+
+
+def test_one_point_duration() -> None:
```
LuisFSegalla, line 122: Move this up

tests/test_specs.py
```diff
@@ -550,59 +605,18 @@ def test_shape(spec: Spec[Any], expected_shape: tuple[int, ...]):
...
-    ):
-        spec.calculate()
-    spec = (  # TODO: refactor when https://github.com/bluesky/scanspec/issues/90
-        Static.duration(0.1) * Line.bounded(DURATION, 0, 0, 2)
-    )
-    frames = spec.calculate()
-    assert len(frames) == 2
-    assert get_constant_duration(frames) is None
+    with pytest.raises(ValueError):
+        spec1.concat(Fly(ConstantDuration(1, spec2)))
```
LuisFSegalla, line 622: Check that the error message match

shihab-dls:
"Anywhere in the tests where gap output has changed, can we put it back, and change the spec so the gap output is the same?"

src/scanspec/specs.py
```diff
@@ -733,7 +833,12 @@ def step(spec: Spec[Axis], duration: float, num: int = 1) -> Spec[Axis | str]:
         spec = step(Line("x", 1, 2, 3), 0.1)
 
     """
-    return spec * Static.duration(duration, num)
+    warnings.warn(
+        f"step method is deprecated! Use {duration} @ spec instead",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    return ConstantDuration(constant_duration=duration, spec=spec)
```
coretl, line 841:
```suggestion
    return duration @ spec
```

tests/test_specs.py
```diff
@@ -550,59 +622,63 @@ def test_shape(spec: Spec[Any], expected_shape: tuple[int, ...]):
...
-def test_frame_with_multiple_axes():
-    spec = Static.duration(0.1).zip(Line.bounded("x", 0, 0, 1))
-    frames = spec.calculate()
-    assert len(frames) == 1
-    assert get_constant_duration(frames) == 0.1
+@pytest.mark.filterwarnings("ignore:fly")
+def test_fly():
+    spec = fly(Line("x", 0, 1, 5), 0.1)
+    (dim,) = spec.calculate()
+    assert dim.midpoints == {x: approx([0, 0.25, 0.5, 0.75, 1])}
```
coretl, line 646: Can we check upper and lower too please

README.md
```diff
@@ -31,11 +31,11 @@ An example ScanSpec of a 2D snaked grid flyscan inside a circle spending 0.4s at
 each point:
 
 ```python
-from scanspec.specs import Line, fly
+from scanspec.specs import Line, Fly, ConstantDuration
```
coretl, line 34:
```suggestion
from scanspec.specs import Line, Fly
```

README.md
```diff
@@ -31,11 +31,11 @@ An example ScanSpec of a 2D snaked grid flyscan inside a circle spending 0.4s at
...
 from scanspec.regions import Circle
 
 grid = Line(y, 2.1, 3.8, 12) * ~Line(x, 0.5, 1.5, 10)
-spec = fly(grid, 0.4) & Circle(x, y, 1.0, 2.8, radius=0.5)
+spec = Fly(ConstantDuration(0.4, grid)) & Circle(x, y, 1.0, 2.8, radius=0.5)
```
coretl, line 38:
```suggestion
spec = Fly(0.4 @ grid) & Circle(x, y, 1.0, 2.8, radius=0.5)
```

docs/how-to/iterate-a-spec.rst
```diff
@@ -89,8 +90,9 @@ You may need to know where there is a gap between points, so that you can do
 something in the turnaround. For example, if we take the x axis of a grid scan,
 you can see it snakes back and forth:
 
->>> from scanspec.specs import Line, fly
->>> grid = fly(Line("y", 0, 1, 2) * ~Line("x", 1, 2, 3), 0.1)
+>>> from scanspec.specs import Line, Fly, ConstantDuration
+>>> spec = Line("y", 0, 1, 2) * ~Line("x", 1, 2, 3)
+>>> grid = Fly(ConstantDuration(0.1,spec))
```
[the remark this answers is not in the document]
LuisFSegalla: Done!

src/scanspec/specs.py
```diff
@@ -714,10 +808,16 @@ def fly(spec: Spec[Axis], duration: float) -> Spec[Axis | str]:
         spec = fly(Line("x", 1, 2, 3), 0.1)
 
     """
-    return spec.zip(Static.duration(duration))
+    warnings.warn(
+        f"fly method is deprecated! Use Fly(ConstantDuration({duration},spec)) instead",
```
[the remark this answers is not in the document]
LuisFSegalla: Done!

src/scanspec/specs.py
LuisFSegalla: Done!

README.md
LuisFSegalla: Done!

LuisFSegalla: Using the `@` operator is now a default way for making a `Spec` with duration, right? Do you want me to updated parts of the code that call `ConstantDuration` explicitly to use `@` as well?

PR #208, opened 2026-04-21
Serialize non-JSON-serializable axes via repr

### LINKED ISSUE #207: Spec.serialize() fails to produce JSON-serializable output when axes are non-primitive objects
When a `Spec` is constructed with a non-primitive axis — such as an `ophyd_async.core.Device` — calling `Spec.serialize()` returns a dict that contains the raw Python object as-is. This makes the output non-JSON-serializable.

Steps to Reproduce

```python
import json
from ophyd_async.core import Device
from scanspec.specs import Fly, Linspace

motor = Device(name="motor")
spec = Fly(Linspace(motor, 1, 2, 5))

serialized = spec.serialize()
json.dumps(serialized)  # raises TypeError: Object of type Device is not JSON serializable
```

hyperrealist:
- Add ophyd_async to the dev dependency group in pyproject.toml
- Update Spec.serialize() to use mode='json' with fallback=repr so that non-JSON-serializable axes (e.g. ophyd_async Device objects) are converted to their repr string, making the output always JSON-serializable
- Add tests for serializing a spec with an ophyd_async Device as axis and verifying the resulting dict is JSON-serializable

Closes #207

### DIFF
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -155,7 +155,7 @@ def test_spec_with_ophyd_async_device_axis_serializes():
     serialized = spec.serialize()
     assert serialized["type"] == "Fly"
     assert serialized["spec"]["type"] == "Linspace"
-    assert serialized["spec"]["axis"] == repr(motor)
+    assert serialized["spec"]["axis"] == motor.name
     assert serialized["spec"]["start"] == 1.0
     assert serialized["spec"]["stop"] == 2.0
     assert serialized["spec"]["num"] == 5
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -177,5 +177,6 @@ def test_spec_with_ophyd_async_device_axis_is_json_serializable():
     motor = Device(name="motor")
     spec = Fly(Linspace(motor, 1, 2, 5))
     serialized = spec.serialize()
+    assert serialized["spec"]["axis"] == motor.name
     # Should not raise
     json.dumps(serialized)
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -2,9 +2,10 @@ from collections.abc import Mapping
 from typing import Any
 
 import pytest
+from ophyd_async.core import Device
 from pydantic import TypeAdapter, ValidationError
 
-from scanspec.specs import Ellipse, Linspace, Spec, Spiral
+from scanspec.specs import Ellipse, Fly, Linspace, Spec, Spiral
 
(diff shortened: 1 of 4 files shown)


tests/test_serialization.py
```diff
@@ -146,3 +147,25 @@ def test_vanilla_serialization():
...
     assert deserialized == ob
+
+
+def test_spec_with_ophyd_async_device_axis_serializes():
+    motor = Device(name="motor")
+    spec = Fly(Linspace(motor, 1, 2, 5))
+    serialized = spec.serialize()
+    assert serialized["type"] == "Fly"
+    assert serialized["spec"]["type"] == "Linspace"
+    assert serialized["spec"]["axis"] == repr(motor)
```
coretl: And what is the repr of a Device?

coretl (commented):
@tpoliaw I know blueapi handles deserialization of json into `Spec[Movable]`. Does it also handle serialization of `Spec[Movable]` into json?

tests/test_serialization.py
hyperrealist: Device repr is something like `"<ophyd_async.core._device.Device object at 0x7f941c9b8590>"`

tpoliaw:
> Does it also handle serialization of `Spec[Movable]` into json?

Not that I'm aware of. On the client side axes are all strings so serialize as themselves and the server never needs to serialize anything.

tests/test_serialization.py
coretl: I think instead of falling back to `repr` we should only support classes with a string `name` attribute, and return its value in serialization.

tests/test_serialization.py
```diff
@@ -146,3 +147,35 @@ def test_vanilla_serialization():
...
+
+
+def test_spec_with_ophyd_async_device_axis_is_json_serializable():
+    import json
+
+    motor = Device(name="motor")
+    spec = Fly(Linspace(motor, 1, 2, 5))
+    serialized = spec.serialize()
+    # Should not raise
+    json.dumps(serialized)
```
coretl, line 182: can you check `serialized["spec"]["axis"]` please

hyperrealist: We also do this test [here](https://github.com/hyperrealist/scanspec/blob/07dbca266e3a17269a1e01963c28e0cfffe7e9ca/tests/test_serialization.py#L158), but I added it again.

coretl (approved):

PR #217, opened 2026-07-06
feat: two-level trigger tree, centred-livetime semantics, and checkpoint pause/resume

hyperrealist:
Implements ADR 0007 (two-level trigger structure + checkpoint pause/resume), which supersedes ADR 0005 and the `positions(TriggerPattern)` part of ADR 0006. Centred-livetime semantics (ADR 0006) are preserved.

- [x] Add ADRs
- [x] Implement ADRs


docs/explanations/decisions/0007-two-level-trigger-structure-and-checkpoint-pause-resume.md
```diff
@@ -0,0 +1,255 @@
...
+Non-empty children on a spacer is a compile-time error — there is no livetime
+window for children to execute within.
+
+### 4. Checkpoint-based pause/resume
+
+**scanspec2's role** is to define the unit of progress and the resume API:
+
+- Each root-level repeat of a `TriggerSequence`'s `trigger_repeat` is a
+  **checkpoint**. `trigger_index` counts completed root-level parent repeats
+  across all `TriggerSequence`s in the window.
```
coretl: Should this include blanks or not?

docs/explanations/decisions/0007-two-level-trigger-structure-and-checkpoint-pause-resume.md
```diff
@@ -0,0 +1,255 @@
...
+
+- `Acquire` → `[frozenset({self.stream_name})]`
+- `Concat` → deduplicated union of each child's list
+- All other wrappers → pass inner value through unchanged
+
+The two axes of PandA resource consumption are orthogonal:
+
+- **Nesting depth** (parent + children in a `TriggerSequence`) determines how
+  many SEQ block levels a single active set requires. This ADR fixes depth at
+  two (parent + one child layer), which fits in a single SEQ block.
```
coretl: One parent plus one child fits in a single SEQ block. Each additional child requires an additional SEQ block

docs/explanations/decisions/0007-two-level-trigger-structure-and-checkpoint-pause-resume.md
```diff
@@ -0,0 +1,255 @@
...
+- **Number of distinct simultaneous active sets** determines how many independent
+  SEQ tables are needed across the full scan. A `Concat` of two differently-named
```
coretl, line 199: independent streams can re-use different outputs of the same SEQ block. There are 6 outputs, so that means up to 6 streams are supported per SEQ block

docs/explanations/decisions/0007-two-level-trigger-structure-and-checkpoint-pause-resume.md
```diff
@@ -0,0 +1,255 @@
...
+### Code changes required
+
+1. **`TriggerRepeat` and `TriggerSequence`** (`core.py`): add the two new
+   dataclasses. Remove `TriggerPattern` and `TriggerGroup`.
+   `Window.trigger_groups` → `Window.trigger_sequences: list[TriggerSequence]`.
+
+2. **`Window.positions()`** (`core.py`): update signature from
+   `float | TriggerPattern` to `float | TriggerRepeat`. When a `TriggerRepeat`
+   is passed, compute trigger instants using the centred-livetime formula for
+   that repeat's `livetime` and `deadtime`.
```
coretl: I'm not convinced this is correct. Let's say I want "all the positions that the Tetramm acquires at". This is a child of a `TriggerSequence`, and there could be a list of `TriggerSequence`s. I also want to be able to specify the position compare points for gated detectors. This means that for a sequence of `N` frames we need to make `N+1` rows that make `N*2` edges. The first row compares at the position at `1/2 deadtime` and makes a high edge, then `N-1` rows that compare at the position at `n*(livetime + deadtime) - 1/2 deadtime` that make a low edge for deadtime then a high edge, then the last row at `N*(livetime + deadtime) - 1/2 deadtime` that makes a low edge. Here's a diagram:
```
windows:  /‾‾‾‾‾‾‾‾‾‾‾‾‾‾\/‾‾‾‾‾‾‾‾‾‾‾‾‾‾\/‾‾‾‾‾‾‾‾‾‾‾‾‾‾\
livetime: __‾‾‾‾‾‾‾‾‾‾‾‾____‾‾‾‾‾‾‾‾‾‾‾‾____‾‾‾‾‾‾‾‾‾‾‾‾__
pcomp:      |           |               |               |  
action:     high        low+high        low+high        low
```
This kind of complexity feels like it should be given to PandA. That's why I suggested we have just a straight `def positions(times)` method that returns the positions at given time intervals, then we push the time chunk generation into the PMAC or PandA logic.

What do you think?

Either way, this diagram needs to make it into the spec as I spent waaay too long drawing it...

docs/explanations/decisions/0007-two-level-trigger-structure-and-checkpoint-pause-resume.md
```diff
@@ -0,0 +1,255 @@
...
+     spacer compile-time error when `children` is non-empty and `livetime=0.0`;
+     `active_stream_sets` covering (a) single `Acquire` → one singleton, (b)
+     `Concat` of two differently-named `Acquire`s → two singletons, (c) `Concat`
+     of two same-named `Acquire`s → one singleton (deduplication).
+
+## Assumptions
+
+### A1 — `_bake_trigger_sequence` always produces a single-entry list; the spacer pattern has no authoring surface yet
+
+`_bake_trigger_sequence` always produces a single-entry `list[TriggerSequence]`
```
coretl: I don't understand this

docs/explanations/decisions/0007-two-level-trigger-structure-and-checkpoint-pause-resume.md
```diff
@@ -0,0 +1,255 @@
...
+produce it. Multi-entry lists only arise via manual `Window` construction
+until a spec-facing authoring surface is added (out of scope for this ADR;
+see PRD §2.5, §11). `_truncate_trigger_sequence` and `_compute_duration` must
+still handle multi-entry lists correctly in all cases, since manually
+constructed `Window`s are a supported input to both.
+
+### A2 — Two nesting levels fit in a single PandA SEQ block
+
+The parent `TriggerSequence.trigger_repeat` and its `children` both encode into a
+single SEQ block — no chained tables are required. The one-child-layer limit in
```
coretl: the one child layering is not so much for SEQ, more for the fact that there must be a single top level stream that we can checkpoint on. Any layers of nesting below that are also expressible collapsed into a second layer

docs/explanations/decisions/0007-two-level-trigger-structure-and-checkpoint-pause-resume.md
```diff
@@ -0,0 +1,255 @@
...
+this ADR is therefore structurally exact, not conservative.
+
+### A3 — SEQ encoding inserts one BITB gate row per root-level parent repeat
+
+The pause/resume guarantee for live rows (max latency = one root-level repeat
+period) requires a `TRIGGER=BITB=1` gate row before each root-level parent repeat.
+The consumer must not collapse N repeats into a single `REPEATS=N` SEQ row — that
+reduces N checkpoints to one. A valid minimal encoding is a two-row sub-table
```
coretl: wrong. The trigger will be checked on every repeat of every row.

docs/explanations/decisions/0007-two-level-trigger-structure-and-checkpoint-pause-resume.md
hyperrealist: Just checking my understanding here. If we don't count a blank we guarantee that a pause landed inside the blank means we overshoots the time gap at resume (because the whole blank re-executes at resume). And this is desirable compared to counting the blank in which case resume doesn't guarantee the minimum time gap (since now it depends on wall time between pause and resume). Since these gaps are minimum requirements (detector readout/recovery time), overshooting is harmless but undershooting isn't, and the simplest way to guarantee the minimum time gap is to simply not count blanks.

API_SPEC.md
```diff
@@ -345,37 +397,41 @@ for window in scan:
...
 
 ```python
 async def run_step_scan(spec: Spec[str, str, str]) -> None:
     scan = spec.compile()
+    assert not scan.has_moving_axes
 
     for window in scan:
         await move(window.static_axes)   # dict[str, float]
 
+        # Single-rate case: one TriggerSequence per detector group, no children.
```
coretl, line 429: Need to assert no children as we can't support them unless we have a triggering system

API_SPEC.md
```diff
@@ -412,17 +471,20 @@ async def run_panda_flyscan(
...
+        # model stops at "trigger N times, this long, this often."
+        tr = seq.trigger_repeat
+        rows += SeqTable.row(
+            repeats=tr.num,
+            trigger=SeqTrigger.IMMEDIATE,
+            time1=int(tr.livetime * 1e6),
+            time2=int(tr.deadtime * 1e6),
+            outa1=True,
+            outa2=False,
+        )
```
coretl, line 490: This needs updating for the panda seq table example with saxs + waxs + panda + tetramm as whiteboarded

API_SPEC.md
```diff
@@ -615,22 +690,42 @@ Operators available on any motion spec node:
...
+otherwise ambiguous, so `Acquire.compile()` raises `ValueError` rather than
+guessing. `detectors` is still required alongside it (it is the arming-time
+description consumed by `WindowedStream.detector_groups`); the two are
+cross-checked for describing the same detector set, not derived from each
+other.
+
 ```python
-# Step scan
+# Step scan — single DetectorGroup, trigger_sequence derived automatically.
 spec: Acquire[str, str, Never] = Acquire(
```
coretl: Check we don't need these annotations, they should be inferred

API_SPEC.md
```diff
@@ -646,13 +741,13 @@ spec: Acquire[str, str, Never] = Acquire(
     ],
 )
 
-# Flyscan — inner axis sweeps continuously; cameras are a ContinuousStream
+# Flyscan — inner axis sweeps continuously; cameras are a ContinuousStream,
+# temperature a free-running MonitorStream. Still a single DetectorGroup.
 spec: Acquire[str, str, str] = Acquire(
```
coretl: ditto

src/scanspec/v2/specs.py
coretl, line 835: Remove livetime and deadtime from DetectorGroup, and make them pure dataclasses
Replace detectors, continuous_streams, monitors, trigger_sequence with:
- Something that takes livetime, deadtime, repeats, etc.
- Is rich enough to make TriggerSequences and DetectorGroups

API_SPEC.md
hyperrealist: image_from_ios

docs/explanations/decisions/0008-unified-trigger-authoring-surface.md
```diff
@@ -0,0 +1,340 @@
...
+detector only needs the total repeat count
+(`exposures_per_collection * collections_per_event`, folded into
+`TriggerRepeat.num`). Reshaping collected data back into events needs the two
+factors kept separate — PRD §5 uses `collections_per_event > 1` to add an
+extra reshaping dimension downstream, and `(2, 3)` vs `(6, 1)` give the same
+trigger count but different output shapes. Because `DetectorGroup` is the
+only place both factors are recorded, `WindowedStream.detector_groups` must
+keep its own `list[DetectorGroup]`, entirely separate from whatever
+`trigger_sequence` the caller supplies — it cannot be derived from the
+already-collapsed `TriggerSequence` after the fact.
```
coretl: wrong rationale, the missing info is how to rewind, not how to construct dimensions

docs/explanations/decisions/0008-unified-trigger-authoring-surface.md
```diff
@@ -0,0 +1,340 @@
...
+
+### 2. `Acquire.trigger_plan` replaces `Acquire.detectors` and `Acquire.trigger_sequence`
+
+```python
+class Acquire(Spec[AxisT, DetectorT, MonitorT]):
+    spec: AnySpec[AxisT, Any, Any]
+    fly: bool = False
+    stream_name: str = "primary"
+    trigger_plan: TriggerPlan[DetectorT] | TriggerGroup[DetectorT] | None = None
+    continuous_streams: Sequence[ContinuousStream[DetectorT]] = ()  # unchanged
```
coretl: trigger_plan and continuous_streams should be merged

docs/explanations/decisions/0008-unified-trigger-authoring-surface.md
```diff
@@ -0,0 +1,340 @@
...
+
+    root/children mirror TriggerSequence's own root/children shape one
+    level up: children fire during every root repeat, in parallel with each
+    other, each at its own integer-multiple rate. Detector sets across root
+    and all children must be disjoint -- checked here structurally at
+    construction time, since timing may still be unresolved. Physical
+    checks (integer-ratio rates, child duration fitting inside the root's
+    livetime, concrete timing) happen at compile() time against the derived
+    TriggerSequence, unchanged from before.
+    """
```
coretl: but how do the integer rates work?

